PR #23, opened 2021-10-13
Update documentation

qwwdfsad:
* Basic KDoc
    * Agent-neutral
    * Updated agent version to avoid workarounds

### DIFF
--- a/src/main/kotlin/kotlinx/kover/KoverTaskExtension.kt
+++ b/src/main/kotlin/kotlinx/kover/KoverTaskExtension.kt
@@ -17,8 +17,7 @@ import java.io.File
  */
 open class KoverTaskExtension(objects: ObjectFactory) {
     /**
-     * Specifies whether the plugin is enabled and
-     * configures test task to collect and generate coverage data.
+     * Specifies whether the plugin is applied to the test task and configures it to collect and generate coverage data.
      */
     public var isEnabled: Boolean = true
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/kotlinx/kover/KoverTaskExtension.kt
```diff
@@ -2,22 +2,72 @@
...
+    /**
+     * Specifies exclusion rules for IntelliJ coverage agent.
+     *
+     * Exclusion rules are represented as set of regular expressions
+     * hat are matched against fully-qualified names of the classes being instrumented.
+     */
+    public var excludes: List<String> = emptyList()
+}
+
+public enum class CoverageAgent {
```
shanshin, line 70: Replace `agent`?

src/main/kotlin/kotlinx/kover/KoverTaskExtension.kt
```diff
@@ -2,22 +2,72 @@
...
+    /**
+     * Specifies inclusion rules for IntelliJ coverage agent.
+     *
+     * Inclusion rules are represented as set of regular expressions
+     * that are matched against fully-qualified names of the classes being instrumented.
+     */
+    public var includes: List<String> = emptyList()
+
+    /**
+     * Specifies exclusion rules for IntelliJ coverage agent.
```
shanshin, line 62: Exclusion rules also apply to JaCoCo.
In this case, they should not be regular expressions, but may contains wildcards `*` and/or `?`.

shanshin (approved):

PR #33, opened 2021-11-01
Multitask

shanshin:


README.md
```diff
@@ -239,24 +172,28 @@ In top level build file
...
+        name = "The project doesn't have to be big"
+        bound {
             maxValue = 100000
             valueType = kotlinx.kover.api.VerificationValueType.COVERED_LINES_COUNT
         }
-        verificationRule {
-            // rule without custom name
+    }
+    rule {
+        // rule without custom name
```
qwwdfsad, line 184:
```suggestion
        // rule without a custom name
```

README.md
```diff
@@ -266,25 +203,55 @@ tasks.test {
 <summary>Groovy</summary>
 
 ```groovy
-tasks.test {
-    kover {
-        verificationRule {
-            name = "The project doesn't have to be big"
+tasks.koverVerify {
+    rule {
+        name = "The project doesn't have to be big"
```
qwwdfsad, line 208: Same here

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=0.3.0
+version=0.3.1
```
qwwdfsad, line 1: With that number of changes is better to make it 0.4.0

src/main/kotlin/kotlinx/kover/tasks/KoverHtmlReportTask.kt
```diff
@@ -0,0 +1,46 @@
...
+            jacocoReport(
+                binaryReportFiles.get(),
+                srcDirs.get(),
+                outputDirs.get(),
+                classpath.get(),
+                null,
+                htmlDirFile,
+            )
+        }
+        project.logger.lifecycle("Kover: HTML report for `${project.name}` file://${htmlDirFile.canonicalPath}/index.html")
```
qwwdfsad, line 44: Better to avoid backticks and use `'` instead

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
```diff
@@ -19,74 +32,192 @@ class KoverPlugin : Plugin<Project> {
...
 
-        doLast {
-            taskExtension.generateXml = taskExtension.generateXml
-                    // turn on XML report for intellij agent if verification rules are defined
-                    || (taskExtension.coverageEngine == CoverageEngine.INTELLIJ && taskExtension.rules.isNotEmpty())
 
-            if (!(taskExtension.isEnabled && (taskExtension.generateXml || taskExtension.generateHtml || taskExtension.rules.isNotEmpty()))) {
-                return@doLast
-            }
+    private fun Project.applyToProject(
```
qwwdfsad, line 68: All tasks have to have a clear description.
Now it seems kinda off-rails

image

qwwdfsad (changes requested):

src/main/kotlin/kotlinx/kover/engines/intellij/IntelliJCoverage.kt
```diff
@@ -124,22 +92,35 @@ private fun readCounterValuesFromXml(file: File): Map<VerificationValueType, Int
 
 
 private fun Task.checkRule(counters: Map<VerificationValueType, Int>, rule: VerificationRule): String? {
-    val minValue = rule.minValue
-    val maxValue = rule.maxValue
+    val boundsViolations = rule.bounds.mapNotNull { it.check(counters) }
+
+    val ruleName = if (rule.name != null) "'${rule.name}' " else ""
+    return if (boundsViolations.size > 1) {
+        "Rule ${ruleName}violated for '${project.name}':" + boundsViolations.joinToString("\n  ", "\n  ")
```
sandwwraith, line 99: Shouldn't there be space?

```suggestion
        "Rule ${ruleName} violated for '${project.name}':" + boundsViolations.joinToString("\n  ", "\n  ")
```

sandwwraith (commented):
Can you tell what changes lead to the fact that Kover now works on Android?

README.md
```diff
@@ -266,25 +203,55 @@ tasks.test {
...
+```
+kover {
+    isEnabled.set(false)
+}
+```
+
+#### Change coverage engine
+```
+kover {
+    coverageEngine.set(kotlinx.kover.api.CoverageEngine.JACOCO)
```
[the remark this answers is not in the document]
shanshin: Yes, now engine defined for all modules, this is done for the future, to support cross-module reports/verifications

README.md
```diff
@@ -79,86 +79,23 @@ buildscript {
...
-
-subprojects {
-    apply(plugin = "kover")
-}
-```
-</details>
 
-<details>
-<summary>Groovy</summary>
+The plugin automatically applied to all modules in the project. It creates tasks `koverReport`, `koverCollectReports`, `koverXmlReport`, `koverHtmlReport`, `koverVerify` and collects coverage during test run.
```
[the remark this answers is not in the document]
shanshin: We need to add this feature later

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
```diff
@@ -19,74 +31,197 @@ class KoverPlugin : Plugin<Project> {
...
+            it.dependsOn(xmlReportTask)
+            it.dependsOn(htmlReportTask)
+        }
+
+        tasks.configureEach {
+            if (it.name == CHECK_TASK_NAME) {
+                it.dependsOn(verificationTask)
+                it.dependsOn(provider {
+                    if (koverExtension.generateReportOnCheck.get()) {
+                        koverReportTask
```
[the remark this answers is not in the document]
shanshin: Now it is so. Maybe we should make two flags.

src/main/kotlin/kotlinx/kover/engines/intellij/IntelliJCoverage.kt
shanshin: Too implicit... because in this case if there is no custom name there will be two spaces between words

shanshin:
> Can you tell what changes lead to the fact that Kover now works on Android?

```
        val enabledTestsProvider = provider {
            tasks.withType(Test::class.java)
                .filter { t -> t.extensions.getByType(KoverTaskExtension::class.java).isEnabled }
        }
        xmlReportTask.dependsOn(enabledTestsProvider)
```
and
```
        val binariesProvider = provider {
            // process binary report only from tasks with enabled cover
            val files = tasks.withType(Test::class.java)
                .map { t -> t.extensions.getByType(KoverTaskExtension::class.java) }
                .filter { e -> e.isEnabled }
                .map { e -> e.binaryReportFile.get() }
            files(files)
        }
        xmlReportTask.binaryReportFiles.set(binariesProvider)
```

Now we do not need to know the exact list of test tasks at the compilation stage - it will be received immediately before the Kover task is executed

src/main/kotlin/kotlinx/kover/engines/intellij/IntelliJCoverage.kt
sandwwraith: It's possible to add some `if`s though

README.md
```diff
@@ -266,25 +271,37 @@ tasks.test {
...
     }
 }
 ```
 </details>
+
+###### Tasks
+The plugin, when applied, automatically creates tasks for the module (all modules, if the project is multi-module, and it applied in root build script):
+- `koverXmlReport` - Generates code coverage XML report for all module's test tasks.
+- `koverHtmlReport` - Generates code coverage HTML report for all module's test tasks.
+- `koverReport` - Generates code coverage HTML and XML reports for all module's test tasks.
```
sandwwraith, line 305: Maybe just 'executes both koverXmlReport and koverHtmlReport tasks'?

sandwwraith:
I think it's good overall, however there are some 'feature regressions': we can't now exclude certain submodules and can't opt-out from XML or HTML report format. If we are OK with that, it's good to go. I'm not sure can exclusion be a blocker to someone or not.

src/main/kotlin/kotlinx/kover/api/KoverExtension.kt
```diff
@@ -8,6 +8,17 @@ import org.gradle.api.model.*
...
+
+    /**
+     * Specifies whether the plugin is applied to the test task and configures it to collect and generate coverage data.
+     */
+    public var isEnabled: Boolean = true
+
+    /**
+     * Specifies the coverage engine to be used to collect execution data.
+     */
+    public val coverageEngine: Property<CoverageEngine> = objects.property(CoverageEngine::class.java)
```
olme04, line 20: haven't you thought about using some sort of sealed class to specify engine+version?
smth like:
```kotlin
sealed class CoverageEngine {
  class IntelliJ(val version: String): CoverageEngine()
  class JaCoCo(val version: String): JaCoCo()
}
```

shanshin:
> I think it's good overall, however there are some 'feature regressions': we can't now exclude certain submodules and can't opt-out from XML or HTML report format. If we are OK with that, it's good to go. I'm not sure can exclusion be a blocker to someone or not.

To disable generating of some report you need to configure task by `enabled = false`. I'll add an example to the documentation

src/main/kotlin/kotlinx/kover/api/KoverExtension.kt
shanshin: Thanks, we will consider this option for the next release.

PR #48, opened 2021-11-08
Ignore empty private constructors of utility classes

### LINKED ISSUE #6: Create an option to not cover private constructors without arguments/default empty constructors
JaCoCo don't include private constructors into report by default (https://github.com/jacoco/jacoco/pull/529).

Intellij have an option to not include private constructors and/or default empty constructors:
image

Kover currently includes all of mentioned above into report. In case of private constructors it causes some additional cases when private constructors are generated implicitly by compiler (for example in sealed classes).

shanshin:
Resolves #6


qwwdfsad (approved):
That's a nice addition! 🔥 

It would be nice to have a test in the future though

PR #55, opened 2021-11-11
Removed implicit dependencies from koverCollectReports task

### LINKED ISSUE #53: Clarify that `koverCollectReports` does not only collect (existing) reports
The docs / task description of `koverCollectReports` reads to me as if only *existing* reports should get collected, whereas executing that task actually *does* also execute the tests (if not up-to-date).

Please clarify whether the current behavior or the docs are correct, and adjust one to the other accordingly.

shanshin:
Resolves #53

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,5 +356,5 @@ The plugin, when applied, automatically creates tasks for the module (all module
 - `koverXmlReport` - Generates code coverage XML report for all module's test tasks.
 - `koverHtmlReport` - Generates code coverage HTML report for all module's test tasks.
 - `koverReport` - Executes both `koverXmlReport` and `koverHtmlReport` tasks.
-- `koverCollectReports` - Collects reports from all submodules in one directory. Main directory is `$buildDir/reports/kover/all`, names for XML reports and dirs for HTML is a projects names. This task does not clear the output directory, but only copies the reports, and the files in it can be overwritten. It is most reliable to call this task only once after `clean` task.
+- `koverCollectReports` - Collects reports from all submodules in one directory. Default directory is `$buildDir/reports/kover/all`, names for XML reports and dirs for HTML is a projects names. Executing this task does not run `koverXmlReport` or `koverHtmlReport`, it only copies previously created reports to the output directory.
 - `koverVerify` - Verifies code coverage metrics based on specified rules. Always executes before `check` task.
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ You may specify custom directory to collect reports from all modules in the buil
 
 ```kotlin
 tasks.koverCollectReports {
-    into { "$buildDir/my-reports-dir" }
+  outputDir.set(layout.buildDirectory.dir("my-reports-dir") )
 }
 ```
 </details>
@@ -244,7 +244,7 @@ tasks.koverCollectReports {
 
 ```groovy
 tasks.koverCollectReports {
-    into "$buildDir/my-reports-dir"
+  outputDir.set(layout.buildDirectory.dir("my-reports-dir") )
 }
 ```
 </details>
@@ -356,5 +356,5 @@ The plugin, when applied, automatically creates tasks for the module (all module
 - `koverXmlReport` - Generates code coverage XML report for all module's test tasks.
 - `koverHtmlReport` - Generates code coverage HTML report for all module's test tasks.
 - `koverReport` - Executes both `koverXmlReport` and `koverHtmlReport` tasks.
-- `koverCollectReports` - Collects reports from all submodules in one directory. Main directory is `$buildDir/reports/kover/all`, names for XML reports and dirs for HTML is a projects names.
+- `koverCollectReports` - Collects reports from all submodules in one directory. Main directory is `$buildDir/reports/kover/all`, names for XML reports and dirs for HTML is a projects names. This task does not clear the output directory, but only copies the reports, and the files in it can be overwritten. It is most reliable to call this task only once after `clean` task.
 - `koverVerify` - Verifies code coverage metrics based on specified rules. Always executes before `check` task.
--- a/README.md
+++ b/README.md
@@ -356,5 +356,5 @@ The plugin, when applied, automatically creates tasks for the module (all module
 - `koverXmlReport` - Generates code coverage XML report for all module's test tasks.
 - `koverHtmlReport` - Generates code coverage HTML report for all module's test tasks.
 - `koverReport` - Executes both `koverXmlReport` and `koverHtmlReport` tasks.
-- `koverCollectReports` - Collects reports from all submodules in one directory. Default directory is `$buildDir/reports/kover/all`, names for XML reports and dirs for HTML is a projects names. Executing this task does not run `koverXmlReport` or `koverHtmlReport`, it only copies previously created reports to the output directory.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -244,7 +244,7 @@ tasks.koverCollectReports {
 
 ```groovy
 tasks.koverCollectReports {
-    into "$buildDir/my-reports-dir"
+  outputDir.set(layout.buildDirectory.dir("my-reports-dir") )
```
sandwwraith, line 247: What's wrong with `$buildDir/my-reports-dir`?

README.md
```diff
@@ -356,5 +356,5 @@ The plugin, when applied, automatically creates tasks for the module (all module
 - `koverXmlReport` - Generates code coverage XML report for all module's test tasks.
 - `koverHtmlReport` - Generates code coverage HTML report for all module's test tasks.
 - `koverReport` - Executes both `koverXmlReport` and `koverHtmlReport` tasks.
-- `koverCollectReports` - Collects reports from all submodules in one directory. Main directory is `$buildDir/reports/kover/all`, names for XML reports and dirs for HTML is a projects names.
+- `koverCollectReports` - Collects reports from all submodules in one directory. Main directory is `$buildDir/reports/kover/all`, names for XML reports and dirs for HTML is a projects names. This task does not clear the output directory, but only copies the reports, and the files in it can be overwritten. It is most reliable to call this task only once after `clean` task.
```
sandwwraith, line 359: This description still does not state whether actual `test` and `koverReport` tasks are invoked or not. 

Also, I don't understand the phrase 'It is most reliable to call this task only once after `clean` task.' What's wrong with invoking `koverCollectReports` after e.g. changing tests? Overwriting old reports seems OK to me. If tests were changed, I get new coverage stats, if they were not, old file still provides correct data.

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
```diff
@@ -43,27 +43,22 @@ class KoverPlugin : Plugin<Project> {
...
     private fun Project.createCollectingTask() {
-        tasks.create(COLLECT_TASK_NAME, Copy::class.java) {
+        tasks.create(COLLECT_TASK_NAME, KoverCollectingTask::class.java) {
             it.group = VERIFICATION_GROUP
             it.description = "Collects reports from all submodules in one directory."
-            allprojects { sub ->
-                val xmlReportTask = sub.tasks.withType(KoverXmlReportTask::class.java).getByName(XML_REPORT_TASK_NAME)
+            it.outputDir.set(project.layout.buildDirectory.dir("reports/kover/all"))
+
+            allprojects { p ->
```
sandwwraith, line 51: `project` instead of `p` looks like a more meaningful name here

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
```diff
@@ -43,27 +43,22 @@ class KoverPlugin : Plugin<Project> {
...
+                val xmlReportTask = p.tasks.withType(KoverXmlReportTask::class.java).getByName(XML_REPORT_TASK_NAME)
                 val htmlReportTask =
-                    sub.tasks.withType(KoverHtmlReportTask::class.java).getByName(HTML_REPORT_TASK_NAME)
+                    p.tasks.withType(KoverHtmlReportTask::class.java).getByName(HTML_REPORT_TASK_NAME)
 
                 it.mustRunAfter(xmlReportTask)
```
sandwwraith, line 56: So it still depends on report tasks and actually invoke tests to collect coverage, right? Then I don't understand what 'remove implicit dependency' means

sandwwraith (changes requested):

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
shanshin: `mustRunAfter` is not dependency, it's ordering rule. It has no effect if the `xmlReportTask` is not scheduled for execution.

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
shanshin: Then the parameter clashes the `Project.project` property - it's confusing.

README.md
shanshin: `outputDir` has type `DirectoryProperty` not `String`

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
shanshin: > Then I don't understand what 'remove implicit dependency' means

The `org.gradle.api.tasks.Copy` task adds an implicit dependency on the task if it uses the output files or directories of that task.

README.md
sandwwraith: Then we probably should clean a particular html directory if we know that we will overwrite the report. Saying 'always invoke clean before using this task' is not a very great user experience. No one will do that anyway, and for big projects consequences of clean may be big (imagine we need to invoke `clean` on Kotlin compiler repo deleting all the artifacts just to re-collect coverage)

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
sandwwraith: `subproject` or at least `proj` then

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
sandwwraith: OK, I see. So now we copy only files that exist if `report` task was invoked, right?

shanshin: We only copy existing files. The task can be executed, but there may be no tests in it or they are all disabled - in this case, the file does not exist either

README.md
sandwwraith: Wouldn't something like this:

```
fun collect() {
... 
     htmlDirs.forEach { (p, d) ->
            outputDir.dir("html/$p").deleteIfExists()
            project.copy {
                it.from(d)
                it.into(outputDir.dir("html/$p"))
            }
     }
```

solve the problem?

README.md
```diff
@@ -356,5 +356,5 @@ The plugin, when applied, automatically creates tasks for the module (all module
 - `koverXmlReport` - Generates code coverage XML report for all module's test tasks.
 - `koverHtmlReport` - Generates code coverage HTML report for all module's test tasks.
 - `koverReport` - Executes both `koverXmlReport` and `koverHtmlReport` tasks.
-- `koverCollectReports` - Collects reports from all submodules in one directory. Main directory is `$buildDir/reports/kover/all`, names for XML reports and dirs for HTML is a projects names.
+- `koverCollectReports` - Collects reports from all submodules in one directory. Default directory is `$buildDir/reports/kover/all`, names for XML reports and dirs for HTML is a projects names. Executing this task does not run `koverXmlReport` or `koverHtmlReport`, it only copies previously created reports to the output directory.
```
sandwwraith, line 359:
```suggestion
- `koverCollectReports` - Collects reports from all submodules in one directory. Default directory is `$buildDir/reports/kover/all`, names for XML reports and dirs for HTML are projects names. Executing this task does not run `koverXmlReport` or `koverHtmlReport`, it only copies previously created reports if they exist to the output directory.
```

sandwwraith (approved):

PR #71, opened 2021-11-23
IntelliJ agent from Maven Central

shanshin:


src/main/kotlin/kotlinx/kover/engines/intellij/IntellijCoverage.kt
```diff
@@ -53,6 +38,47 @@ internal fun Task.intellijReport(
     }
 }
 
+private fun Writer.writeModuleReportJson(
```
qwwdfsad, line 41: Please add a comment, e.g. with a resulting JSON and its purpose for easier maintainability in the future

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
```diff
@@ -28,10 +27,6 @@ class KoverPlugin : Plugin<Project> {
     private val defaultJacocoVersion = "0.8.7"
 
     override fun apply(target: Project) {
-        target.repositories.maven {
```
qwwdfsad, line 31: Can't we add Maven Central here?

For the people who use company-wide mirrors, we can provide an option to opt-out, but it will work like a charm for everyone else. WDYT?

shanshin: I'm not sure if implicitly adding a repository is a good idea. Users are not notified about this in any way, and thus some may have a hidden problem with dependencies.

Of course, it is convenient for most users not to do additional actions, but at the same time, the vast majority of users have already added mavenCentral to the list of repositories. For example, in build.gradle.ts files should use the `maven Central' repository, because the `kotlin-scripting-compiler-embeddable` artifact is loaded from there.

qwwdfsad: Thanks for the rationale! Let's use your approach then

README.md
```diff
@@ -29,6 +30,31 @@ Minimal supported `Gradle` version: `6.4`.
 * Customizable filters for instrumented classes
 
 ## Quickstart
+### Using IntelliJ coverage engine
```
qwwdfsad, line 33: I'd suggest moving it to the very bottom as it's not really relevant for quickstart and should not bother 95% of our users

shanshin: OK

qwwdfsad (approved):

PR #75, opened 2021-11-26
Fixed caching of Kover report tasks

### LINKED ISSUE #68: Kover report tasks not cacheable
When running a simple project twice with build cache enabled, we can see that the Kover report task is skipped:

image

which leads to no report being generated nor taken from a cache.
The way I would expect it working would be:
- First run stores the report into the build cache
- Second run does not skip the report task, just takes the report from the cache

shanshin:
Resolves #68


qwwdfsad:
It definitely would be nice to have a test on it

src/main/kotlin/kotlinx/kover/api/KoverExtension.kt
```diff
@@ -6,33 +6,39 @@ package kotlinx.kover.api
...
      * Specifies the version of JaCoCo dependency.
      */
     val jacocoEngineVersion: Property<String> = objects.property(String::class.java)
+        @Input get
 
     /**
      * Specifies whether the reports will be generated within 'check' task execution.
      */
     val generateReportOnCheck: Property<Boolean> = objects.property(Boolean::class.java)
+        @Input get
```
qwwdfsad, line 41: here and everywhere: it's enough to mark the field with the `@Input`/`@Output`/`@Nested` without the need to introduce getter separately, isn't it?

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
```diff
@@ -148,6 +150,24 @@ class KoverPlugin : Plugin<Project> {
...
+                .map { t -> t.extensions.getByType(KoverTaskExtension::class.java) }
+                .filter { e -> e.isEnabled }
+                .map { e -> e.smapFile.orNull }
+                /*
+                 Binary reports and SMAP files have same ordering for IntelliJ engine:
+                    * SMAP file is null if coverage engine is a JaCoCo by default - in this case property is unused
+                    * SMAP file not creates by JaCoCo - property is unused
+                    * test task have no sources - in this case binary report and SMAP file not exists
+                 */
+                .filter { f -> f?.exists() == true }
```
qwwdfsad, line 164: super nit -- when working with nulls, `f?.exists() ?: true` can be more readab;e

src/main/kotlin/kotlinx/kover/engines/intellij/IntellijCoverage.kt
```diff
@@ -95,8 +97,9 @@ private fun Writer.writeModuleReportJson(
     appendLine("""  "modules": [""")
     appendLine("""    { "reports": [ """)
 
+    val smapIterator = smapFiles.iterator()
     appendLine(binaryReportFiles.joinToString(",\n        ", "        ") { f ->
-        """{"ic": "${f.safePath()}", "smap": "${f.safePath()}.smap"}"""
+        """{"ic": "${f.safePath()}", "smap": "${smapIterator.next().safePath()}"}"""
```
qwwdfsad, line 102: You can remove `smapIterator` and just use `smapIterator.first()`

qwwdfsad (approved):

src/main/kotlin/kotlinx/kover/api/KoverExtension.kt
shanshin: Gradle searches for annotations only on getters.
Documentation to the annotation: `... Annotations on setters or just the field in Java are ignored.`

src/main/kotlin/kotlinx/kover/engines/intellij/IntellijCoverage.kt
shanshin: Unfortunately, `first` will not be enough - it is necessary to read all the elements of the iterator inside a loop strictly in order

qwwdfsad: Oh, I see, it's captured in lambda.

shanshin:
> It definitely would be nice to have a test on it

Tests added

src/main/kotlin/kotlinx/kover/api/KoverExtension.kt
shanshin: Changed to `@get:Nested` etc

PR #86, opened 2021-12-03
Unified agent filters

### LINKED ISSUE #21: Unify agents filters
Jacoco uses wildcard like `*` in filters for the agent, IntelliJ agent uses regular expressions.
This makes it difficult to switch between agents. It is necessary to design a solution to this problem or document it and leave it as is.

shanshin:
All filters for both engines now use only `*` and `?` wildcards.
Resolves #21


src/functionalTest/kotlin/kotlinx/kover/test/functional/core/Writer.kt
```diff
@@ -0,0 +1,186 @@
...
+    return ""
+}
+
+private fun ProjectBuilderState.buildSettings(slice: ProjectSlice): String {
+    return loadSettingsTemplate(slice)
+        .replace("//SUBMODULES", buildSubmodulesIncludes(submodules.keys))
+}
+
+private fun ModuleBuilderState.buildScripts(slice: ProjectSlice): String {
+//        val scripts = if (language == GradleScriptLanguage.KOTLIN) kotlinScripts else groovyScripts
```
sandwwraith, line 162: It seems that this function is not needed anymore?

src/main/kotlin/kotlinx/kover/api/KoverTaskExtension.kt
```diff
@@ -37,33 +37,19 @@ open class KoverTaskExtension(objects: ObjectFactory) {
...
-     * ### Inclusion rules for IntelliJ
-     *
-     * Inclusion rules are represented as set of regular expressions
-     * that are matched against fully-qualified names of the classes being instrumented.
-     *
-     * ### Inclusion rules for JaCoCo
-     *
-     * Inclusion rules are represented as set of [JaCoCo-specific](https://www.eclemma.org/jacoco/trunk/doc/report-mojo.html#includes)
-     * fully qualified name that also supports `*` and `?`.
+     * Inclusion rules are represented as set of fully-qualified names of the classes being instrumented
```
sandwwraith, line 42:
```suggestion
     * Inclusion rules are represented as a set of fully-qualified names of the classes being instrumented.
```

src/main/kotlin/kotlinx/kover/api/KoverTaskExtension.kt
```diff
@@ -37,33 +37,19 @@ open class KoverTaskExtension(objects: ObjectFactory) {
...
+     * that also supports `*` and `?` wildcards.
```
sandwwraith, line 43:
```suggestion
     * It's possible to use `*` and `?` wildcards.
```

src/main/kotlin/kotlinx/kover/api/KoverTaskExtension.kt
```diff
@@ -37,33 +37,19 @@ open class KoverTaskExtension(objects: ObjectFactory) {
...
-     *
-     * Exclusion rules are represented as set of regular expressions
-     * that are matched against fully-qualified names of the classes being instrumented.
-     *
-     * ### Inclusion rules for JaCoCo
+     * Specifies class exclusion rules for coverage engine.
      *
-     * Exclusion rules are represented as set of [JaCoCo-specific](https://www.eclemma.org/jacoco/trunk/doc/report-mojo.html#excludes)
-     * fully qualified name that also supports `*` and `?`.
+     * Inclusion rules are represented as set of fully-qualified names of the classes being instrumented
```
sandwwraith, line 51:
```suggestion
     * Exclusion rules are represented as a set of fully-qualified names of the classes being instrumented.
```

src/main/kotlin/kotlinx/kover/api/KoverTaskExtension.kt
```diff
@@ -37,33 +37,19 @@ open class KoverTaskExtension(objects: ObjectFactory) {
...
+     * that also supports `*` and `?` wildcards.
```
sandwwraith, line 52:
```suggestion
     * It's possible to use `*` and `?` wildcards.
```

src/main/kotlin/kotlinx/kover/engines/intellij/IntellijAgent.kt
```diff
@@ -55,20 +54,37 @@ internal class IntellijAgent(val config: Configuration) {
...
-            extension.excludes.forEach { i ->
-                pw.appendLine(i)
+            extension.excludes.forEach { e ->
+                pw.appendLine(e.replaceWildcards())
             }
         }
     }
+
+    private fun String.replaceWildcards(): String {
+        val builder = StringBuilder(length * 2)
```
sandwwraith, line 71: Most of the chars won't result in two chars, so it is logical to place `StringBuilder(length)` here

src/main/kotlin/kotlinx/kover/engines/jacoco/JacocoAgent.kt
```diff
@@ -29,17 +28,24 @@ internal class JacocoAgent(val config: Configuration, private val project: Proje
...
-            "jmx=false"
+            "jmx=false",
+            extension.includes.filterString("includes"),
+            extension.excludes.filterString("excludes")
         ).joinToString(",")
     }
 }
 
+private fun List<String>.filterString(name: String): String? {
+    if (isEmpty()) return null
```
sandwwraith, line 45: `isBlank` or `isEmpty`?

src/functionalTest/kotlin/kotlinx/kover/test/functional/cases/InstrumentationFilteringTests.kt
```diff
@@ -0,0 +1,53 @@
+package kotlinx.kover.test.functional.cases
+
+import kotlinx.kover.api.*
+import kotlinx.kover.test.functional.cases.utils.*
+import kotlinx.kover.test.functional.core.*
+import kotlinx.kover.test.functional.core.BaseGradleScriptTest
+import kotlin.test.*
+
+internal class InstrumentationFilteringTests : BaseGradleScriptTest() {
```
sandwwraith, line 9: Please also add tests for inclusion and/or subsetting, e.g.

```
excludes = ["pkg.*Example*"]
includes = ["pkg.FooExample*"]
```

sandwwraith (changes requested):

src/functionalTest/kotlin/kotlinx/kover/test/functional/core/Runner.kt
```diff
@@ -1,326 +1,108 @@
...
 
-private data class ProjectSlice(val language: GradleScriptLanguage, val type: ProjectType, val engine: CoverageEngine)
 
-private class SubmoduleBuilderImpl : ModuleBuilder<SubmoduleBuilderImpl> {
-    override fun verification(rules: Iterable<VerificationRule>): SubmoduleBuilderImpl {
-        TODO("Not yet implemented")
-    }
+private class XmlReportImpl(file: File) : XmlReport {
+    private val document = DocumentBuilderFactory.newInstance()
+        .also { it.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false) }
```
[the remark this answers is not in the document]
shanshin: The option with this name disables checking the dtd file, because the JaCoCo report has the `<!DOCTYPE report PUBLIC "-//JACOCO//DTD Report 1.1//EN" "report.dtd">`

src/functionalTest/kotlin/kotlinx/kover/test/functional/core/Writer.kt
```diff
@@ -0,0 +1,186 @@
...
+import kotlinx.kover.api.*
+import java.io.*
+
+private const val TEMPLATES_PATH = "src/functionalTest/templates"
+private const val BUILD_SCRIPTS_PATH = "$TEMPLATES_PATH/scripts/buildscripts"
+private const val SETTINGS_PATH = "$TEMPLATES_PATH/scripts/settings"
+private const val SOURCES_PATH = "$TEMPLATES_PATH/sources"
+
+internal fun ProjectBuilderState.createProject(rootDir: File, slice: ProjectSlice): File {
+    val projectDir = File(rootDir, System.currentTimeMillis().toString()).also { it.mkdirs() }
```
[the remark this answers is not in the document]
shanshin: if you create a directory with the `Files.createTempFile`, then you will need to monitor its deletion yourself, here I use a directory that is deleted by JUnit itself.

src/main/kotlin/kotlinx/kover/engines/intellij/IntellijAgent.kt
shanshin: It seems to me that filter are most often used with a wildcard *, or a dot  - in this case an escape symbol will be added, the size will overflow and the capacity will be doubled.

src/main/kotlin/kotlinx/kover/engines/jacoco/JacocoAgent.kt
shanshin: In this case, I'm working with a `List`, not a `String`. 
If no filters are defined (an empty list), we return null so that it is not passed to the agent arguments.

src/functionalTest/kotlin/kotlinx/kover/test/functional/core/Writer.kt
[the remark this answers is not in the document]
shanshin: `currentTimeMillis` is unstable so I replaced it with a string generated from a slice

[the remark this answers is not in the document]
sandwwraith: You can create a temp file in any directory you want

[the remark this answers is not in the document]
shanshin: I think the encoded slice is more obvious than a random number

src/functionalTest/kotlin/kotlinx/kover/test/functional/core/Runner.kt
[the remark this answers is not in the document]
sandwwraith: Then it worth commenting, I think

src/functionalTest/kotlin/kotlinx/kover/test/functional/cases/InstrumentationFilteringTests.kt
```diff
@@ -0,0 +1,58 @@
...
+            .languages(GradleScriptLanguage.KOTLIN, GradleScriptLanguage.GROOVY)
+            .types(ProjectType.KOTLIN_JVM, ProjectType.KOTLIN_MULTIPLATFORM)
+            .engines(CoverageEngine.INTELLIJ, CoverageEngine.JACOCO)
+            .sources("simple")
+            .configTest(
+                """includes = listOf("org.jetbrains.*Cla?s")""",
+                """includes = ['org.jetbrains.*Cla?s']"""
+            )
+            .configTest(
+                """excludes = listOf("org.jetbrains.*Exa?ple*")""",
```
sandwwraith, line 45: It's worth documenting somewhere that exclusion filters have priority over inclusion ones

sandwwraith (approved):

src/functionalTest/kotlin/kotlinx/kover/test/functional/cases/InstrumentationFilteringTests.kt
shanshin: Added comment to extension's docs

Issue #43, opened 2021-11-05
Add a combined report for a multi-module build


altavir:
It would be useful to have a single combined result dashboard for the whole project consisting of multiple subprojects.

shanshin:
Cross-module reports design is in process (#20).
It would be great if you write use cases from your project.

altavir:
Our test coverage is not very good right now and it would be a good task to gradually improve it. But we have a lot of modules and it would be good to understand where to focus first. Looking into a dashboard for all modules, we can decide if we want to focus on some modules first or if we need to raise coverage gradually in all modules. 

Also one needs to take into account #16 or any other shield-generating tool. It usually provides a single number for the project.

mustafaozhan:
So the use case for me, I have multimodule project also, it is KMP project but for the android side it shouldn't really matter, I would like to have a single report.xml file combined with all the modules in app.

shanshin:
Implemented in `0.5.0-RC`.

PR #90, opened 2021-12-09
Implemented aggregate reports

### LINKED ISSUE #20: Design multimodule projects support

shanshin:
Resolves #20, #43


christian-draeger:
Great news :) was waiting for this actual feature so much. 
When it will be released? 
Is it possible to test / use it even before official release?

sandwwraith (commented):
It's rather list of questions than an actual review, but we have to start from something

src/functionalTest/kotlin/kotlinx/kover/test/functional/core/Builder.kt
```diff
@@ -100,26 +97,32 @@ private open class ModuleBuilderImpl<B : ModuleBuilder<B>>(val moduleState: Modu
...
     }
 
     override fun config(kotlin: String, groovy: String): B {
-        moduleState.kotlinScripts += kotlin
-        moduleState.groovyScripts += groovy
+        moduleState.testScripts += GradleScript(kotlin, groovy)
+        return this as B
+    }
+
+    override fun dependency(script: String): B {
```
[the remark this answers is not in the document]
shanshin: Dependency syntax is most often identical in Kotlin and Groovy.

src/functionalTest/kotlin/kotlinx/kover/test/functional/core/Writer.kt
```diff
@@ -173,7 +190,7 @@ private fun loadSettingsTemplate(slice: ProjectSlice): String {
 }
 
 private fun loadScriptTemplate(root: Boolean, slice: ProjectSlice): String {
-    val filename = if (root) "root" else "child"
+    val filename = if (root) "root" else "submodule"
```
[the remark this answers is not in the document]
shanshin: There can be an unlimited number of submodules, because this file is copied. However, it is assumed that they are all of the first level.

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
```diff
@@ -30,173 +40,191 @@ class KoverPlugin : Plugin<Project> {
...
+        tasks.create(MODULE_REPORT_TASK_NAME) {
+            it.group = VERIFICATION_GROUP
+            it.dependsOn(xmlReportTask)
+            it.dependsOn(htmlReportTask)
+            it.description = "Generates code coverage HTML and XML reports for all enabled test tasks in one module."
+        }
+
+        tasks.configureEach {
+            if (it.name == CHECK_TASK_NAME) {
+                it.dependsOn(verifyTask)
```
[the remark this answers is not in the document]
shanshin: I think in most cases, by default, it is enough to generate only an aggregated report for all modules. It seems redundant to generate a report separately for each module each time, if necessary, the user can explicitly call `koverModuleReport`.

src/main/kotlin/kotlinx/kover/tasks/KoverXmlReport.kt
```diff
@@ -0,0 +1,69 @@
...
+import org.gradle.api.tasks.*
+
+@CacheableTask
+open class KoverXmlReportTask : KoverAggregateTask() {
+
+    /**
+     * Specifies file path of generated XML report file with coverage data.
+     */
+    val xmlReportFile: RegularFileProperty = project.objects.fileProperty()
+        @OutputFile get
```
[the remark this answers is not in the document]
shanshin: didn't notice (

src/main/kotlin/kotlinx/kover/tasks/KoverVerification.kt
```diff
@@ -0,0 +1,137 @@
...
+    public fun rule(configureRule: Action<VerificationRule>) {
+        rulesInternal += project.objects.newInstance(VerificationRuleImpl::class.java, project.objects)
+            .also { configureRule.execute(it) }
+    }
+
+    @TaskAction
+    fun verify() {
+        verify(report(), coverageEngine.get(), rulesInternal, classpath.get()) {
+            val xmlReport =
+                this.project.tasks.withType(KoverXmlReportTask::class.java)
```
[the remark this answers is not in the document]
shanshin: I tried to move it into a common function, but the difficulty is that `KoverXmlReportTask` and `KoverXmlModuleReportTask` do not have a suitable common parent. 
At the same time, it seemed redundant to me to make a high hierarchy of classes and it was easier to copy code that should soon be deleted than to overload classes with inheritance.

src/functionalTest/kotlin/kotlinx/kover/test/functional/cases/utils/Defaults.kt
```diff
@@ -1,14 +1,24 @@
...
 
-internal const val DEFAULT_JACOCO_KJVM_BINARY = "kover/test.exec"
-internal const val DEFAULT_JACOCO_KMP_BINARY = "kover/jvmTest.exec"
+internal fun defaultSmapFile(projectType: ProjectType): String {
+    return defaultBinaryReport(CoverageEngine.INTELLIJ, projectType) + ".smap"
+}
 
-internal const val DEFAULT_XML = "reports/kover/report.xml"
-internal const val DEFAULT_HTML = "reports/kover/html"
+internal fun defaultXmlReport() = "reports/kover/report.xml"
```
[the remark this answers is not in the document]
shanshin: This is done because at the moment there is no certainty that additional parameters will not be needed to get the default value (for example, as in the case of defaultBinaryReport). So it turns out the general style that getting all the values is a functions, and not a mixture of `val` and `fun`.

However, I do not insist on such a decision, if `val` is perceived better, I can change it.

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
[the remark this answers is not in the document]
sandwwraith: What if check is called only for submodule? E.g. `:mysubmodule:check`. Will the report be generated and verification rules apply?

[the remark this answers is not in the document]
shanshin: When executing such a command, no reports are generated, but the verification rules will be checked if they are declared for the module.

README.md
```diff
@@ -10,14 +10,14 @@ Minimal supported `Gradle` version: `6.4`.
 ## Table of content
 - [Features](#features)
 - [Quickstart](#quickstart)
-  - [Apply plugin to single-module project](#apply-plugin-to-single-module-project)
+  - [Apply plugin to single-module build](#apply-plugin-to-single-module-build)
```
sandwwraith, line 13:
```suggestion
  - [Apply plugin to a single-module build](#apply-plugin-to-single-module-build)
```

README.md
```diff
@@ -10,14 +10,14 @@ Minimal supported `Gradle` version: `6.4`.
...
     - [Applying plugins with the plugins DSL](#applying-plugins-with-the-plugins-dsl)
     - [Legacy Plugin Application: applying plugins with the buildscript block](#legacy-plugin-application-applying-plugins-with-the-buildscript-block)
-  - [Apply plugin to multi-module project](#apply-plugin-to-multi-module-project)
+  - [Apply plugin to multi-module build](#apply-plugin-to-multi-module-build)
```
sandwwraith, line 16:
```suggestion
  - [Apply plugin to a multi-module build](#apply-plugin-to-multi-module-build)
```

README.md
```diff
@@ -32,7 +32,7 @@ Minimal supported `Gradle` version: `6.4`.
 * Customizable filters for instrumented classes
 
 ## Quickstart
-### Apply plugin to single-module project
+### Apply plugin to single-module build
```
sandwwraith, line 35: (don't forget to update table of contents)
```suggestion
### Apply plugin to a single-module build
```

README.md
```diff
@@ -94,12 +94,8 @@ apply plugin: 'kover'
 ```
 </details>
 
-### Apply plugin to multi-module project
-To apply the plugin to all modules in the project, you need to apply the plugin only to the root module, as shown [above](#apply-plugin-to-single-module-project).
-
-**There are no dependencies between tasks from different modules, they are executed independently.**
-
-**Cross-module tests are not supported in reports and validation yet. For each test, only the classpath belonging to the current module is taken.**
+### Apply plugin to multi-module build
```
sandwwraith, line 97:
```suggestion
### Apply plugin to a multi-module build
```

README.md
```diff
@@ -94,12 +94,8 @@ apply plugin: 'kover'
...
+To apply the plugin to all modules, you just need to apply the plugin only to the root module, as shown [above](#apply-plugin-to-single-module-build).
```
sandwwraith, line 98: I'd add here something like 'Applying the plugin to submodules if you have already applied it to the root module will cause configuration errors'

sandwwraith, line 98:
```suggestion
To apply the plugin to all Gradle modules, you just need to apply the plugin only to the root module, as shown [above](#apply-plugin-to-single-module-build).
```

README.md
```diff
@@ -194,21 +190,22 @@ android {
 </details>
 
 
-### Configuring reports
-If you need to change the name of the XML report file or HTML directory, you may configure the corresponding tasks
+### Configuring aggregated reports
+If you need to change the name of the XML report file or HTML directory, you may configure the corresponding tasks in 
+the module in which the plugin is applied (usually this is the root module).
```
sandwwraith, line 195: I think we need to add here some kind of definition for an aggregated report. E.g. 'aggregated report provides report using combined classpath from all subprojects and available when plugin is applied to the root project' or something like that

README.md
```diff
@@ -354,19 +391,58 @@ tasks.koverVerify {
...
-- `koverXmlReport` - Generates code coverage XML report for all module's test tasks.
-- `koverHtmlReport` - Generates code coverage HTML report for all module's test tasks.
-- `koverReport` - Executes both `koverXmlReport` and `koverHtmlReport` tasks.
-- `koverCollectReports` - Collects reports from all submodules in one directory. Default directory is `$buildDir/reports/kover/all`, names for XML reports and dirs for HTML are projects names. Executing this task does not run `koverXmlReport` or `koverHtmlReport`, it only copies previously created reports if they exist to the output directory.
-- `koverVerify` - Verifies code coverage metrics based on specified rules. Always executes before `check` task.
+The plugin, when applied, automatically creates tasks for the module in which it is applied (usually this is the root module):
+- `koverHtmlReport` - Generates code coverage HTML report for all enabled test tasks in all modules.
+- `koverXmlReport` - Generates code coverage XML report for all enabled test tasks in all modules. 
+- `koverReport` - Executes both `koverXmlReport` and `koverHtmlReport` tasks.  Executes before `check` task if `generateReportOnCheck` is `true`.
```
sandwwraith, line 431: It's not specified what `generateReportOnCheck` is. Is it a gradle property? Or Kover extension property? Probably we shouldn't mention it at all.

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
```diff
@@ -212,60 +242,67 @@ class KoverPlugin : Plugin<Project> {
...
 
         return taskExtension
     }
+
+    private fun Project.checkAlreadyApplied() {
+        var parent = parent
+
+        while (parent != null) {
+            if (parent.plugins.hasPlugin(KoverPlugin::class.java)) {
+                throw GradleException("Kover plugin already applied: in module '${parent.name}' and child module '${this.name}'")
```
sandwwraith, line 274:
```suggestion
                throw GradleException("Kover plugin is applied in both parent module '${parent.name}' and child module '${this.name}'. Kover plugin should be applied only in parent module.")
```

sandwwraith (approved):

PR #102, opened 2021-12-21
Upgraded IntelliJ Agent version and added copying of error logs

shanshin:


sandwwraith (commented):
How a user will understand that the error happened during intellij agent work? Will build fail? If so, we need to add some kind of message 'check for logs in kover/errors/...'

shanshin:
> How a user will understand that the error happened during intellij agent work? Will build fail? If so, we need to add some kind of message 'check for logs in kover/errors/...'

The build will not fall, but I am not sure that without additional learning of the reasons for the appearance of this file, it is need crushing the build.
The files contain the agent's internal messages, so studying them will not help the user to understand if there is a problem in his project.

However, in standard cases, this file should definitely not appear, so functional tests have been added.

sandwwraith:
So we just copying the file in case we want to ask user to attach it to the issue report. OK, make sense

sandwwraith (approved):

PR #103, opened 2021-12-21
Kotlin Multiplatform plugin adapter rewritten to use reflection

### LINKED ISSUE #100: Kotlin Multiplatform report is empty
Repro: https://github.com/Reproducers/kover-repro

Report:

```xml
<?xml version="1.0" ?>
<report name="Intellij Coverage Report">
<counter type="INSTRUCTION" missed="0" covered="0"/>
<counter type="BRANCH" missed="0" covered="0"/>
<counter type="LINE" missed="0" covered="0"/>
<counter type="METHOD" missed="0" covered="0"/>
<counter type="CLASS" missed="0" covered="0"/>
</report>
```

In the same repo, there is a JVM library that is working.

Both engines fail.

If I apply the plugin directly to the multiplatform subproject instead of in the root project, it works.

If I try to add kover to all subprojects, the report is empty in the multiplatform project.

```kotlin
buildscript {
    repositories {
        mavenCentral()
    }

    dependencies {
        classpath("org.jetbrains.kotlinx:kover:0.4.4")
    }
}

subprojects {
    pluginManager.apply("org.jetbrains.kotlinx.kover")
}
```

It is really weird behavior.

At the same time, everything I have tried is working in the JVM library.

shanshin:
Fixes #100

### DIFF
--- a/src/main/kotlin/kotlinx/kover/adapters/KotlinMultiplatformPluginAdapter.kt
+++ b/src/main/kotlin/kotlinx/kover/adapters/KotlinMultiplatformPluginAdapter.kt
@@ -4,8 +4,11 @@
 
 package kotlinx.kover.adapters
 
+import groovy.lang.*
 import kotlinx.kover.adapters.api.*
 import org.gradle.api.*
+import org.gradle.api.file.*
+import org.gradle.internal.metaobject.*
 import org.jetbrains.kotlin.gradle.dsl.*
 import org.jetbrains.kotlin.gradle.plugin.*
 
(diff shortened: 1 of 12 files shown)


src/main/kotlin/kotlinx/kover/adapters/KotlinMultiplatformPluginAdapter.kt
```diff
@@ -4,29 +4,43 @@
...
 
+    @Suppress("UNCHECKED_CAST")
     override fun findDirs(project: Project): PluginDirs {
         return safe(project) {
             this.plugins.findPlugin("kotlin-multiplatform") ?: return@safe PluginDirs(emptyList(), emptyList())
 
-            val extension = project.extensions.findByType(
-                KotlinMultiplatformExtension::class.java
-            ) ?: return@safe PluginDirs(emptyList(), emptyList())
+            val extension =
```
qwwdfsad, line 20: This part of the trickery definitely needs some explanation with comments for the sake of future maintainability

sandwwraith: +1

sandwwraith (approved):

src/main/kotlin/kotlinx/kover/adapters/KotlinMultiplatformPluginAdapter.kt
shanshin: @qwwdfsad, @sandwwraith In addition to the comment, I can move reflection to the fail safe path, because in most cases the Kover will be declared in the same project as the kotlin Multiplatform Plugin.

It will be a copy paste, but it will work faster and there will be an example of exactly how the code should work with the reflection.

qwwdfsad: Yeah, that would be nice

PR #105, opened 2021-12-21
Inverted the flags for disabling instrumentation of test tasks

shanshin:


sandwwraith (commented):
Please describe in commit message why this change was needed

sandwwraith (approved):

shanshin:
> Please describe in commit message why this change was needed

OK.
In practice, the name 'IsEnabled` is misleading in that it seems that if you wrote 'IsEnabled = true' instrumentation should work anyway.

However, if you configure the project like this
```
kover {
    isEnabled = false
}

tasks.test {
    extensions.configure(kotlinx.kover.api.KoverTaskExtension::class) {
        isEnabled = true
    }
}
```

then the test task will not be instrumented anyway, despite the fact that we have explicitly enabled it.

`isDisabled` guarantees that if you assigned it `true`, then the task instrumentation will definitely not happen. However, this does not guarantee instrumentation if the value is 'false`

PR #109, opened 2021-12-23
Release 0.5.0-RC

shanshin:


CHANGELOG.md
```diff
@@ -1,3 +1,25 @@
+0.5.0-RC / 2021-12-24
+===================
+
+This is a release candidate for the next version. In this version, an aggregated report on all projects has been added, 
+and the plugin API has been significantly redesigned. So we ask you to evaluate it and share your feedback on whether 
+the API has become more convenient.
+
+### Features
+* Unified coverage agents filters. Now only the characters '*' or '?' are used as wildcards for both IntelliJ and JaCoCo agents. 
+**Regular expressions are no longer supported by the IntelliJ agent as filters of instrumented classes.** (#21)
```
sandwwraith, line 10: Is this supposed to be a separate list item? If not, consider removing line break and joining with previous line. It would become long, yes, but it's better than to guess 'was it supposed to be `* **Reg...` or `**Reg...`

sandwwraith (approved):

PR #118, opened 2022-01-12
Add version catalog for dependency management

jsoberg:
Implemented gradle version catalogs for easier/safer dependency usage.

I also added `local.properties` to the `.gitignore` for those who use Android Studio.

Ran tests for verification, all 9 passed.


JavierSegoviaCordoba:
I think Version Catalogs should be used instead of buildSrc approach which is not cache friendly.

jsoberg:
I could add that in instead of this, however I don't think we could reference the Kotlin version from TOML when we set the 'kotlin("jvm")' plugin

JavierSegoviaCordoba:
Version catalog supports plugin aliases too

jsoberg:
> Version catalog supports plugin aliases too

True I forgot about that option, I'd be happy to move this over to version catalogs if that's best. I wasn't aware that using `buildSrc` wasn't cache-friendly

jsoberg:
Merged in main and resolved conflicts, I just needed to update to 1.6.10 in the version catalog to mirror that update.

jsoberg:
Let me know if you'd like to add this functionality and I'll open up another PR

shanshin:
Hi, thanks for the contribution.
I propose to implement this feature when it is stabilized in Grade (it won't be a preview) and there will be full support in the IDE.

shanshin:
@JavierSegoviaCordoba, does version `7.4` still require a `enableFeaturePreview("VERSION_CATALOGS")` call?

JavierSegoviaCordoba:
@shanshin no, it can be removed

PR #126, opened 2022-01-20
Added support of parallel tests execution

### LINKED ISSUE #113: Code coverage is incorrect
[This project](https://github.com/JavierSegoviaCordoba/kotlin-stdlib) should have `Collections.kt` and `Strings.kt` partially/totally covered but Kover is having a few weird behaviors.

Checking the HTML files, not sure if XML ones are correct.

- At the start, in Windows machine, `Strings.kt` had 0% and `Collections.kt` had 100%. I did a `--rerun-tasks` plus `--no-build-cache` and now `Strings.kt` is covered but `Collections.kt` has no percentage.
- In macOS I am having the same behavior as the second one in the Windows machine. The project didn't exist in the machine, it was a new clone.
- In CI (macOS + Ubuntu + Windows) I am getting what I think is the correct behavior, both, `Strings.kt` and `Collections.kt` being covered. All code coverage reports can be downloaded at the end of [this link](https://github.com/JavierSegoviaCordoba/kotlin-stdlib/actions/runs/1647430250).

shanshin:
Fixes #113

### DIFF
--- a/src/main/kotlin/kotlinx/kover/KoverPlugin.kt
+++ b/src/main/kotlin/kotlinx/kover/KoverPlugin.kt
@@ -102,7 +102,7 @@ class KoverPlugin : Plugin<Project> {
         }
 
         tasks.withType(Test::class.java).configureEach { t ->
-            t.configTest(providers, agents)
+            t.configTestTask(providers, agents)
         }
     }
 
@@ -238,7 +238,7 @@ class KoverPlugin : Plugin<Project> {
         return extension
     }
 
-    private fun Test.configTest(
+    private fun Test.configTestTask(
         providers: AllProviders,
         agents: Map<CoverageEngine, CoverageAgent>
     ) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/kotlinx/kover/KoverPlugin.kt
```diff
@@ -278,7 +279,34 @@ class KoverPlugin : Plugin<Project> {
     }
 }
 
-private class IntellijErrorLogChecker(private val taskExtension: KoverTaskExtension) : Action<Task> {
+/*
+  To support parallel tests, both Coverage Engines work in append to data file mode.
```
sandwwraith, line 283: Hm, I'm not sure one can append to file in parallel without causing it to be a mess. Are you sure it won't be a problem?

sandwwraith (approved):

PR #159, opened 2022-03-30
Add support for branches and instructions metric in verify task

mskonovalov:
Adding other types of coverage metric verification: branches and instructions


shanshin:
Thanks for the contribution, but unfortunately we have decided to follow a different path here: verification will be performed not on the plugin side, but by the IntelliJ Coverage Engine.
In the next versions of the plugin, this code will be removed and the IntelliJ verifier call will be implemented.

mskonovalov:
Ok, thanks for the information.
Do you mean it will still be available via plugin?
So it will be available for CI?

In this case would we add the current implementation until there is an alternative?

I'm trying to migrate the project to Gradle and this is unfortunately one of the important functions.
This will force me to maintain a fork until the new implementation is available

shanshin:
@mskonovalov, yes, verification will also be configuring via plugin, only the implementation of the verification itself will change. The current implementation is not a very nice temporary solution.

I expect that the next release will happen with a new verification implementation, so it will be pointless to apply these changes to the main branch. Therefore, unfortunately, you will have to use the forked version until the next release.

PR #163, opened 2022-04-13
Implemented the use of the full path to the project

### LINKED ISSUE #151: Incorrect gradle project analysis on multi-module projects when subprojects have the same name
I'm playing with Kover (v0.5.0) these days and I noticed a weird behaviour on multi-module projects with a nested hierarchy. When there are submodules with the same name, Kover often analyses the wrong project.

Given the following project structure:
```
:feature
  :first
  :another-nesting
     :first
     :second
  :nested
     :first
     :second
```

`./gradlew :feature:first:koverHtmlReport` is analyzing `:feature:nested:first`

Screenshot 2022-03-11 at 11 53 00

`./gradlew :feature:nested:first:koverHtmlReport` is correct

Screenshot 2022-03-11 at 11 53 30

`./gradlew :feature:another-nesting:first:koverHtmlReport` is analyzing again `:feature:nested:first`

Screenshot 2022-03-11 at 11 53 53

-----------------------

Zero configuration from my side, just applied the plugin. 

Downgraded to `v0.4.4` and everything works fine. 

Everything should be reproducible in this repo https://github.com/ologe/kover-testing

shanshin:
Fixes #151

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ kover {
     intellijEngineVersion.set("1.0.656")    // change version of IntelliJ agent and reporter
     jacocoEngineVersion.set("0.8.7")        // change version of JaCoCo agent and reporter
     generateReportOnCheck = true            // false to do not execute `koverMergedReport` task before `check` task
-    disabledProjects = setOf()              // setOf("project-name") to disable coverage for project with name `project-name`
+    disabledProjects = setOf()              // setOf("project-name") or setOf(":project-name") to disable coverage for project with path `:project-name` (`:` for the root project)
     instrumentAndroidPackage = false        // true to instrument packages `android.*` and `com.android.*`
     runAllTestsForProjectTask = false       // true to run all tests in all projects if `koverHtmlReport`, `koverXmlReport`, `koverReport`, `koverVerify` or `check` tasks executed on some project
 }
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -332,7 +332,7 @@ kover {
     intellijEngineVersion.set("1.0.656")    // change version of IntelliJ agent and reporter
     jacocoEngineVersion.set("0.8.7")        // change version of JaCoCo agent and reporter
     generateReportOnCheck = true            // false to do not execute `koverMergedReport` task before `check` task
-    disabledProjects = setOf()              // setOf("project-name") to disable coverage for project with name `project-name`
+    disabledProjects = setOf()              // setOf(":project-name") to disable coverage for project with path `:project-name` (`:` for the root project)
```
qwwdfsad, line 335: It seems like an additional burden (apart from the fact that in **minor** release we'll break every configured project) for users to put `:` before the project name.

I suggest another approach:

* Allow using **both** short names and FQ names
* When using short name, search for projects with the same short name and refuse to configure the plugin unless unambigous name is specified

qwwdfsad (changes requested):
Also, it would be nice to have a test for that

README.md
shanshin: implemented

shanshin:
> Also, it would be nice to have a test for that

`MultiProjectTests#testDisableSubproject` and `MultiProjectTests#testDisableSubprojectByPath`

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
```diff
@@ -302,7 +306,7 @@ private class BinaryReportCleanupAction(
 
         if (!taskExtension.isDisabled
             && !koverExtension.isDisabled
-            && !koverExtension.disabledProjects.contains(projectName)
+            && !koverExtension.disabledProjectsPaths.contains(projectPath)
```
qwwdfsad, line 309: I see that extension now has both `disabledProjectsPaths` and `disabledProjects` with the same documentation. Is it intended? If so, they should clearly be distinct by their documentation

qwwdfsad (changes requested):

src/main/kotlin/kotlinx/kover/KoverPlugin.kt
shanshin: Documentation on `disabledProjectsPaths` is not required, because this field is not available to the user.
But I will leave a correct dev comment.

qwwdfsad (approved):
Nice 👍

PR #171, opened 2022-04-28
introduce GitHub Actions

nise-nabe:
add Check and Gradle Wrapper Validation workflow.

* https://github.com/marketplace/actions/gradle-build-action
  * distribution is same with other Kotlin org code
    * https://github.com/search?l=YAML&q=org%3AKotlin+distribution&type=Code
* https://github.com/marketplace/actions/gradle-wrapper-validation
  * https://docs.gradle.org/current/userguide/gradle_wrapper.html#wrapper_checksum_verification

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,25 @@
+name: Check
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


nise-nabe:
It seems to be used TeamCity already.

PR #176, opened 2022-05-07
Update KoverExtension.kt

anshuman35:


shanshin:
Hi,
thanks for the contribution, but this is breaking changes. 
Also, the JaCoCo version cannot be represented as an int value, because it is represented as `0.8.8`

PR #198, opened 2022-08-01
Release 0.6.0-BETA

shanshin:


CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
+0.6.0-BETA / 2022-08-02
+===================
+In this version, the plugin API has been fully redesigned. The new API allows you to configure Kover in a more flexible manner, there is no need to configure Kover or test tasks separately.
```
qwwdfsad, line 3: Please leave a link to a migration guide. Also, please ensure that link from deprecation message leads here

gradle.properties
```diff
@@ -1,5 +1,5 @@
 version=0.6.0-SNAPSHOT
```
qwwdfsad, line 1: `0.6.0-Beta-SNAPSHOT`.

Also, note how we are using `Beta` instead of `BETA` all over the Kotlin, let's stick to the single naming scheme

qwwdfsad (changes requested):

CHANGELOG.md
shanshin: https://github.com/Kotlin/kotlinx-kover/blob/v0.6.0-Beta/docs/migration-to-0.6.0.md accessible (temporary tag leads to this PR, then it will be placed on release commit)

qwwdfsad (approved):

PR #204, opened 2022-08-12
Added instrumentation filtering by common filters

### LINKED ISSUE #201: Jacoco class filter does not work on kover 0.6.0-Beta
With Intellij engine 1.0.675 kover properly setups class filtering and the resulting report is good.

But for the same configuration, the Jacoco engine 0.8.8 does not filter classes at all. 

Notes
- Project - Android multi module with different build configs and different flavors, but the tests were performed for a single variant using kover instrumentation and task exclusion
- Described problem reproduces in a single module and also in merged report
- example filters:
`com.companyxyz.package.*`
`*.di.*`

shanshin:
Fixes #201

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,8 @@ Examples `my.package.ClassName` or `my.*.*Name` are allowed, while `my/package/C
 
 Exclusion rules have priority over inclusion ones.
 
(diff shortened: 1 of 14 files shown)


qwwdfsad (approved):
👍

Issue #212, opened 2022-08-17
Instructions count in verify task is different from the one displayed in report


AleksDanil:
0.6.0-Beta

You can check the attached simple project for reproduce.
Basically when I select instruction as counter in `verify` task numbers which I see in error differs from the html report.
For example in attached project report states branches coverage as `65.2% (30/46)` and from the verification errors I see that covered branches count is 20, and missed count is 12. Which also gives different percentage.
[CounterTypeIssue.zip](https://github.com/Kotlin/kotlinx-kover/files/9365578/CounterTypeIssue.zip)

shanshin:
https://youtrack.jetbrains.com/issue/IDEA-300309/Incorrect-values-of-instruction-counter-in-the-verifier

zuevmaxim:
Fixed in agent 1.0.680

shanshin:
Fixed via f7f0d1c6ef8f8bc4b0096fd8a4275749d4e76f03

Issue #210, opened 2022-08-17
java.lang.IndexOutOfBoundsException on running `verify` task with `VerificationTarget` other than `ALL`


AleksDanil:
0.6.0-Beta

```
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
        at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
        at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
        at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
        at kotlinx.kover.engines.intellij.IntellijVerificationKt.raiseViolations(IntellijVerification.kt:271)
        at kotlinx.kover.engines.intellij.IntellijVerificationKt.intellijVerification(IntellijVerification.kt:48)
        at kotlinx.kover.engines.commons.EngineManager.verify(EngineManager.kt:59)
        at kotlinx.kover.tasks.KoverVerificationTask.verify(KoverVerificationTask.kt:22)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
        at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:104)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.doExecute(StandardTaskAction.java:58)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:51)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:29)
        at org.gradle.api.internal.tasks.execution.TaskExecution$3.run(TaskExecution.java:242)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:29)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:26)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:66)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:59)
        at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:157)
[...]

PR #214, opened 2022-08-22
Fixed instrumentation counter in IntelliJ verifier

### LINKED ISSUE #211: `Verify` task error says about "lines" when counterType is Branch or Instruction
0.6.0-Beta

Set counter for verify task INSTRUCTION or BRANCH:
```
verify {
            ...
                bound {
                    minValue = 20
                    maxValue = 20
                    counter = kotlinx.kover.api.CounterType.BRANCH 
                    valueType = kotlinx.kover.api.VerificationValueType.COVERED_COUNT
                }
            }
```
In case of error it will always be about lines:
`Rule 'test' violated: missed lines percentage is 75.000000, but expected maximum is 20`
Expected:
`Rule 'test' violated: missed instructions/branches percentage is 75.000000, but expected maximum is 20`

shanshin:
- IntelliJ Engine version upgraded to 1.0.680
- added tests on verification error

Fixes #211, #210, #212

### DIFF
--- a/src/functionalTest/kotlin/kotlinx/kover/test/functional/cases/VerificationTests.kt
+++ b/src/functionalTest/kotlin/kotlinx/kover/test/functional/cases/VerificationTests.kt
@@ -38,15 +38,15 @@ internal class VerificationTests : BaseGradleScriptTest() {
     }
 
     @Test
-    fun testNotVerifiedIntelliJ() {
-        val build = diverseBuild(languages = ALL_LANGUAGES, engines = listOf(CoverageEngineVendor.INTELLIJ))
+    fun testVerificationError() {
+        val build = diverseBuild(languages = ALL_LANGUAGES, engines = ALL_ENGINES)
         build.addKoverRootProject {
-            sourcesFrom("simple")
+            sourcesFrom("verification")
 
             kover {
                 verify {
                     rule {
-                        name = "test rule"
+                        name = "counts rule"
                         bound {
                             minValue = 58
                             maxValue = 60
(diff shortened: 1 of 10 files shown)


src/functionalTest/kotlin/kotlinx/kover/test/functional/cases/VerificationTests.kt
```diff
@@ -57,55 +57,93 @@ internal class VerificationTests : BaseGradleScriptTest() {
...
-            output {
-                assertTrue {
-                    this.contains(
-                        "[ant:jacocoReport] Rule violated for bundle :: lines covered ratio is 0.50, but expected minimum is 0.58\n" +
-                                "[ant:jacocoReport] Rule violated for bundle :: lines covered count is 4, but expected maximum is 3"
-                    )
-                }
+        build.prepare().runWithError("koverHtmlReport", "koverVerify") {
+            verification {
+                assertIntelliJResult("""Rule 'counts rule' violated:
```
qwwdfsad, line 101: Works for me, but in case when there is more than one such test, I would suggest extracting test data (these long strings) into `test-resources` text files available by `@TestName` similarly to this:

qwwdfsad (approved):

src/functionalTest/kotlin/kotlinx/kover/test/functional/cases/VerificationTests.kt
shanshin: I think yes, with an increase in the number of verification tests, these lines should be put into files.

PR #215, opened 2022-08-22
Updated docs for onCheck properties

### LINKED ISSUE #213: with `onCheck.set(true)` merged tasks are not launched
0.6.0-Beta

I configured simple multi module project and set `onCheck.set(true)` for `verify`, `htmlReport`, `xmlReport` inside `koverMerged` for parent module:
```kotlin
koverMerged {
    enable()

    xmlReport {
        onCheck.set(true)
    }
    
    htmlReport {
        onCheck.set(true)
    }

    verify {
        onCheck.set(true)
        ...
    }

}
```
None of koverMergedHtmlReport, koverMergedXmlReport, koverMergeVerify is launched when I run check task.

You can check attached project [MultiModule.zip](https://github.com/Kotlin/kotlinx-kover/files/9377495/MultiModule.zip)

shanshin:
Resolves #213

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,7 +248,7 @@ kover {
     }
 
     xmlReport {
-        onCheck.set(false) // true to run koverXmlReport task during the execution of the check task
+        onCheck.set(false) // true to run koverXmlReport task during the execution of the check task (if it exists) of the current project
         reportFile.set(layout.buildDirectory.file("my-project-report/result.xml")) // change report file name
         overrideFilters { 
             classes { // override common class filter
@@ -259,7 +259,7 @@ kover {
     }
 
     htmlReport {
-        onCheck.set(false) // true to run koverHtmlReport task during the execution of the check task
+        onCheck.set(false) // true to run koverHtmlReport task during the execution of the check task (if it exists) of the current project
         reportDir.set(layout.buildDirectory.dir("my-project-report/html-result")) // change report directory
         overrideFilters { 
             classes { // override common class filter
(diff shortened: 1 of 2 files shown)


qwwdfsad (changes requested):
It seems to be an incomplete fix -- it's easy to misunderstand this documentation and still misuse API.

I suggest throwing a configuration error if `onCheck` is explicitly set for projects without `check` task

shanshin:
> I suggest throwing a configuration error if onCheck is explicitly set for projects without check ta

For this checking with it is necessary to slightly change the API - the default value will now be "no value".

qwwdfsad:
>For this checking with it is necessary to slightly change the API - the default value will now be "no value".

It is not, because `no value` and `false` are identical semantically anyway.
Our goal here is to provide fail-fast mechanism for the common scenario where user misconfigures their project by explicitly invoking `onCheck(true)`.

It is perfectly fine not to fail on `onCheck(false)` as this behaviour is indistinguishable from setting nothing at all

shanshin:
1. as far as I know, it is not possible to add a listener to `Property.set(true)`
2. verify tasks true by default now
3. when checking, I will not distinguish the default `true` from the one set explicitly
4. if I replace the `onCheck` property with a function with an immediate check for the presence of the `check` task, then this may work unstable, because the task can be created after assigning a value (eq in `afterEvaluate`)

qwwdfsad (approved):
:okayface:

PR #216, opened 2022-08-22
Ordered report tasks before verification

### LINKED ISSUE #209: koverXmlReport task is not launched with check task when both verify and koverXmlReport are `onCheck.set(true)`
Simple project is configured like this:
```kotlin
plugins {
    kotlin("jvm") version "1.7.10"
    id("org.jetbrains.kotlinx.kover") version "0.6.0-Beta"
}

kover {
    xmlReport {
        onCheck.set(true)
    }

        verify {
            onCheck.set(true)
            rule { 
                isEnabled = true 
                name = null 
                target =
                    kotlinx.kover.api.VerificationTarget.ALL

                bound {
                    minValue = 10
                    maxValue = 20
                    counter =
                        kotlinx.kover.api.CounterType.LINE
                    valueType =
                        kotlinx.kover.api.VerificationValueType.COVERED_PERCENTAGE 
                }
            }
        }
    }
}
```

Run `check` task. koverXmlReport task is not launched, xml report is not generated.
With `onCheck.set(false)` for verify task is launched and report is generated.

No such issue for `htmlReport`. Also could not reproduce on 0.5.1 with similar setup.

shanshin:
Resolves #209

### DIFF
--- a/src/main/kotlin/kotlinx/kover/appliers/KoverMergedApplier.kt
+++ b/src/main/kotlin/kotlinx/kover/appliers/KoverMergedApplier.kt
@@ -92,6 +92,8 @@ private class ProcessMergeExtensionAction(private val extension: KoverMergedConf
         // TODO `onlyIf` block moved out from config lambda because of bug in Kotlin compiler - it implicitly adds closure on `Project` inside onlyIf's lambda
         verifyTask.onlyIf { extension.verify.hasActiveRules() }
 
+        // ordering of task calls, so that if a verification error occurs, reports are generated and the values can be viewed in them
+        verifyTask.shouldRunAfter(xmlTask, htmlTask)
 
         container.tasks.create(MERGED_REPORT_TASK_NAME) {
             it.group = VERIFICATION_GROUP
--- a/src/functionalTest/kotlin/kotlinx/kover/test/functional/cases/TasksOrderingTests.kt
+++ b/src/functionalTest/kotlin/kotlinx/kover/test/functional/cases/TasksOrderingTests.kt
@@ -1,0 +1,54 @@
+package kotlinx.kover.test.functional.cases
+
+import kotlinx.kover.test.functional.cases.utils.*
+import kotlinx.kover.test.functional.core.*
+import kotlinx.kover.test.functional.core.BaseGradleScriptTest
+import kotlin.test.*
+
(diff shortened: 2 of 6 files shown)


src/main/kotlin/kotlinx/kover/appliers/KoverMergedApplier.kt
```diff
@@ -92,6 +92,8 @@ private class ProcessMergeExtensionAction(private val extension: KoverMergedConf
         // TODO `onlyIf` block moved out from config lambda because of bug in Kotlin compiler - it implicitly adds closure on `Project` inside onlyIf's lambda
         verifyTask.onlyIf { extension.verify.hasActiveRules() }
 
+        // ordering of task calls, so that if a verification error occurs, reports are generated and the values can be viewed in them
+        verifyTask.shouldRunAfter(xmlTask, htmlTask)
```
qwwdfsad, line 96: Does it fix the situation when a unit test in `check` fails and a report is not generated?

If so, a test for that would be nice

shanshin: Yes, #209  is just about that

shanshin: Unfortunately, if the test has fallen, then no, this is a Gradle limitation.
The report generation task depends on the test task, if any of the tasks failed, then Gradle stops executing all dependent tasks

qwwdfsad: Is it true for `--continue` as well?

shanshin: According [docs](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:continue_build_on_failure) `If a task fails, any subsequent tasks that were depending on it will not be executed.` - because when generating reports, the output of test tasks is used, so even in this case they will not be executed

qwwdfsad: Can you still add a test on the change though?

shanshin: Ready

qwwdfsad (approved):

PR #219, opened 2022-08-24
Added example projects

shanshin:


src/main/kotlin/kotlinx/kover/util/Versions.kt
```diff
@@ -0,0 +1,40 @@
...
+
+internal class SemVer(val major: Int, val minor: Int, val patch: Int): Comparable<SemVer> {
+    companion object {
+        fun ofThreePartOrNull(version: String): SemVer? {
+            val parts = version.split(".")
+            if (parts.size != 3) return null
+
+            val major = parts[0].toIntOrNull()?: return null
+            val minor = parts[1].toIntOrNull()?: return null
+            val patch = parts[2].toIntOrNull()?: return null
```
sandwwraith, line 15: What about smthing like `1.2.3-SNAPSHOT`? I think it may be a case when someone wants to try locally-built agents.

src/functionalTest/kotlin/kotlinx/kover/test/functional/framework/writer/KoverWriter.kt
```diff
@@ -0,0 +1,136 @@
...
+import kotlinx.kover.api.*
+import kotlinx.kover.test.functional.framework.common.*
+import kotlinx.kover.test.functional.framework.configurator.*
+import kotlinx.kover.test.functional.framework.configurator.TestVerifyConfig
+
+internal fun ScriptAppender.writeKover(kover: TestKoverConfig?) {
+    if (kover == null) {
+        if (engineForced != null) {
+            block("kover") {
+                writeEngine(null)
```
sandwwraith, line 16: I think it's better to drop a comment for why this `if` is needed, because first you check `engineForced != null` and then call `writeEngine(null)` — looks like a typo, maybe `writeEngine(engineForced)` was meant to be used?

src/functionalTest/kotlin/kotlinx/kover/test/functional/framework/writer/ScriptAppender.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    private fun indent(count: Int): String {
+        return when (count) {
+            0 -> ""
+            1 -> "    "
+            2 -> "        "
+            3 -> "            "
+            4 -> "                "
+            5 -> "                    "
+            else -> "    ".repeat(count)
```
sandwwraith, line 75: Why not use `repeat(count)` without `when`?

src/functionalTest/kotlin/kotlinx/kover/test/functional/framework/writer/ScriptAppender.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    @Suppress("NOTHING_TO_INLINE")
+    inline fun lineIf(shouldWrite: Boolean, line: String) {
+        if (shouldWrite) line(line)
+    }
+
+    inline fun block(name: String, shouldWrite: Boolean = true, content: ScriptAppender.() -> Unit) {
+        if (!shouldWrite) return
+        line("$name {")
+        stepInto().content()
```
sandwwraith, line 50: It may be easier to do just `indents++/indents--` instead of creating a new anonymous object each time

examples/defaults/build.gradle.kts
```diff
@@ -0,0 +1,134 @@
...
+    mavenCentral()
+}
+
+dependencies {
+    testImplementation(kotlin("test"))
+}
+
+kover {
+    isDisabled.set(false) // true to disable instrumentation and all Kover tasks in this project
+    engine.set(kotlinx.kover.api.DefaultIntellijEngine) // change Coverage Engine
```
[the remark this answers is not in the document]
shanshin: Yes, it is possible to use specify custom version of engine e.g. `kotlinx.kover.api.IntellijEngine("x.x.xxx")`

examples/merged/build.gradle.kts
```diff
@@ -0,0 +1,38 @@
...
+    enable()
+
+    filters {
+        classes {
+            includes += "kotlinx.kover.examples.merged.*"
+            excludes += listOf("kotlinx.kover.examples.merged.utils.*", "kotlinx.kover.examples.merged.subproject.utils.*")
+        }
+
+        projects {
+            excludes += "excluded"
```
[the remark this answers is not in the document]
shanshin: Now it is expected that a plugin for should be applied in all subprojects, otherwise it should be explicitly excluded from reports.

This was specially done, in case the plugin was applied at the root project, users should be notified (with error) that they need to apply the plugin in subprojects by themselves, and not wait for the Kover itself to do it.

gradle.properties
```diff
@@ -1,3 +1,4 @@
+#latest release version with "-SNAPSHOT" prefix
```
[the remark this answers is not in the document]
shanshin: Unfortunately, now it is not always possible to know in advance what the version of the next release will be.

This is done to check the version of the latest release in the examples. It might be worth adding an additional property.

src/functionalTest/kotlin/kotlinx/kover/test/functional/framework/checker/Checker.kt
```diff
@@ -0,0 +1,449 @@
+/*
```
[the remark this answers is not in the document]
shanshin: not ok (

src/functionalTest/kotlin/kotlinx/kover/test/functional/framework/checker/Checker.kt
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+}
+
+private class CheckerContextImpl(
+    val rootDir: File,
+    val result: BuildResult,
+    val path: String
+) : CheckerContext {
+    val projectDir: File = rootDir.sub(path.removeSuffix(":").replace(':', '/'));
```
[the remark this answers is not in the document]
shanshin: It is quite difficult to implement this, because you need to use the Groovy and Kotlin parser - now it is unjustified, because there are no examples that would use such an opportunity.

src/functionalTest/kotlin/kotlinx/kover/test/functional/framework/starter/Example.kt
```diff
@@ -0,0 +1,103 @@
...
+            val name = it.name
+            it.isDirectory && !excludes.contains(name) && (includes.isEmpty() || includes.contains(name))
+        }?.toList() ?: emptyList()
+
+        return files.stream().map { ExampleArgs(it, commands) }
+    }
+}
+
+private class ExampleArgs(private val exampleDir: File, private val commands: List<String>) : Arguments {
+    override fun get(): Array<Any> {
```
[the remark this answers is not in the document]
shanshin: Unfortunately, it will not be possible to transfer this above, because the `CheckerContext` should already get into the arguments of the function, for argument we need to prepare a directory.

Judging by the logs, the arguments are calculated once before the test starts.

src/main/kotlin/kotlinx/kover/util/Versions.kt
shanshin: I haven't decided how to compare the `-SNAPSHOT` version and the regular one, so at the moment I decided to leave it that way.
In any case, there is now a requirement for the minimum version of the agent.

src/functionalTest/kotlin/kotlinx/kover/test/functional/framework/writer/ScriptAppender.kt
shanshin: allocation optimization (possibly redundant)

src/functionalTest/kotlin/kotlinx/kover/test/functional/framework/writer/ScriptAppender.kt
shanshin: Yep, thanks!

examples/defaults/build.gradle.kts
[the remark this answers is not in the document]
sandwwraith: If this example is meant to be read as example to external users, then it's better to add smth like `// to change engine, use `IntellijEngine(xxx) or JacocoEngine(xxx)` or `// to change engine, follow instructions here [...]`

examples/merged/build.gradle.kts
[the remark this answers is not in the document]
sandwwraith: IMO it's not very convenient when I have a lot of (lets say 100) subprojects and want to apply Kover to half of them — as far as I understand, I need to apply Kover to all 100 and then add 50 of them to excludes?

Anyway, it's out of scope of this PR, so let's discuss this later

src/main/kotlin/kotlinx/kover/util/Versions.kt
sandwwraith: For now, you can just use smth like `.removeSuffix("-SNAPSHOT")` or `.removeAfter("-")`

src/functionalTest/kotlin/kotlinx/kover/test/functional/framework/common/Environment.kt
```diff
@@ -0,0 +1,34 @@
...
+
+package kotlinx.kover.test.functional.framework.common
+
+import java.io.*
+
+internal val koverVersion = System.getProperty("koverVersion")
+    ?: throw Exception("System property 'koverVersion' not defined for functional tests")
+
+internal val recentKoverVersion = System.getProperty("recentKoverVersion")
+    ?: throw Exception("System property 'koverVersion' not defined for functional tests")
```
sandwwraith, line 13:
```suggestion
    ?: throw Exception("System property 'recentKoverVersion' not defined for functional tests")
```

src/functionalTest/kotlin/kotlinx/kover/test/functional/framework/writer/ScriptAppender.kt
sandwwraith: I'd rather remove it to avoid confusion, it's not that performance-critical

shanshin: It's better to add a comment why it is so

sandwwraith: OK

sandwwraith: Please add a comment then

sandwwraith (approved):

Issue #222, opened 2022-08-26
Error "Kover plugin not applied in projects" when there are nested subproject


aSemy:
If I have a Gradle project where there are subprojects nested inside empty directories, then Gradle considers these directories to be subprojects, but they are not.

```
.
├── my-gradle-project/
│   └── subprojects/
│       ├── alpha-project/
│       │   └── build.gradle.kts
│       └── beta-project/
│           └── build.gradle.kts
├── build.gradle.kts
└── settings.gradle.kts
```

Here Gradle considers `:subprojects` to be a subproject, and so does Kover, but it should be ignored.

This can be set manually, but it would be nicer if this wasn't necessary. I think the solution to this is to use variant aware sharing: https://docs.gradle.org/7.5.1/userguide/cross_project_publications.html

Here's a demo from Gradle: https://docs.gradle.org/7.3/samples/sample_jvm_multi_project_with_code_coverage.html

shanshin:
Hi,
could you please show the full error message?

aSemy:
```
16:45:48: Executing 'koverMergedHtmlReport --stacktrace'...

Configuration cache is an incubating feature.
Calculating task graph as no configuration cache is available for tasks: koverMergedHtmlReport
Type-safe project accessors is an incubating feature.
> Task :buildSrc:generateExternalPluginSpecBuilders UP-TO-DATE
> Task :buildSrc:extractPrecompiledScriptPluginPlugins UP-TO-DATE
> Task :buildSrc:compilePluginsBlocks UP-TO-DATE
> Task :buildSrc:generatePrecompiledScriptPluginAccessors UP-TO-DATE
> Task :buildSrc:generateScriptPluginAdapters UP-TO-DATE
> Task :buildSrc:compileKotlin UP-TO-DATE
> Task :buildSrc:compileJava NO-SOURCE
> Task :buildSrc:compileGroovy NO-SOURCE
> Task :buildSrc:pluginDescriptors UP-TO-DATE
> Task :buildSrc:processResources UP-TO-DATE
> Task :buildSrc:classes UP-TO-DATE
> Task :buildSrc:inspectClassesForKotlinIC UP-TO-DATE
> Task :buildSrc:jar UP-TO-DATE
> Task :buildSrc:assemble UP-TO-DATE
> Task :buildSrc:compileTestKotlin NO-SOURCE
> Task :buildSrc:pluginUnderTestMetadata UP-TO-DATE
[...]

Issue #221, opened 2022-08-26
Failed to apply plugin 'org.jetbrains.kotlinx.kover', Cannot run Project.afterEvaluate(Action) when the project is already evaluated


aSemy:
* Kover 0.6.0
* Kotlin 1.7.10
* Gradle 7.5.1

Please see this project at this commit https://github.com/aSemy/skrape.it/tree/6bc29c3d27f59c58fa56367c910feea371405a3e

`./gradlew koverMergedReport`

It's related to 

* applying the Kover plugin in a buildSrc script
* this `afterEvaluate` block 
   https://github.com/Kotlin/kotlinx-kover/blob/0fcc584f3ee3fdac6886cc807196350b10bbc503/src/main/kotlin/kotlinx/kover/appliers/KoverMergedApplier.kt#L27

```
An exception occurred applying plugin request [id: 'org.jetbrains.kotlinx.kover']
...
Caused by: org.gradle.api.internal.plugins.PluginApplicationException: Failed to apply plugin 'org.jetbrains.kotlinx.kover'.
...
Caused by: org.gradle.api.InvalidUserCodeException: Cannot run Project.afterEvaluate(Action) when the project is already evaluated.
	at org.gradle.api.internal.project.DefaultProject.failAfterProjectIsEvaluated(DefaultProject.java:1068)
	at org.gradle.api.internal.project.DefaultProject.afterEvaluate(DefaultProject.java:1049)
	at kotlinx.kover.appliers.KoverMergedApplierKt.applyMerged(KoverMergedApplier.kt:27)
```

<details>
  <summary><b>Click here to see the full log</b></summary>
  

```log
15:16:03: Executing 'koverMergedReport'...

Configuration cache is an incubating feature.
Calculating task graph as no configuration cache is available for tasks: koverMergedReport
Type-safe project accessors is an incubating feature.
> Task :buildSrc:generateExternalPluginSpecBuilders UP-TO-DATE
> Task :buildSrc:extractPrecompiledScriptPluginPlugins
> Task :buildSrc:compilePluginsBlocks

> Task :buildSrc:generatePrecompiledScriptPluginAccessors
Failed to generate type-safe Gradle model accessors for the following precompiled script plugins:
 - src/main/kotlin/buildsrc/convention/kotlin-jvm.gradle.kts

[...]

PR #223, opened 2022-08-26
update project to use kotlin-dsl plugin

aSemy:
- update methods to use Gradle Kotlin DSL
- use embeddedKotlinVersion to help align with Gradle embedded Kotlin version
- add simple smoke test for testing if the plugin works

This is an initial step in fixing #222  and #221, and improving compatibility with Gradle config cache


src/test/kotlin/KoverPluginTest.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import java.nio.file.Files
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+
+class KoverPluginTest {
```
aSemy, line 11: I found the existing tests pretty difficult to understand and unpick. I created this small test so I could more easily adjust the test to fetch the output build log.

src/main/kotlin/kotlinx/kover/engines/intellij/IntellijVerification.kt
```diff
@@ -28,19 +27,19 @@ internal fun Task.intellijVerification(
...
     val verifyRequest = File(temporaryDir, "verify-request.json")
     val verifyResponseFile = File(temporaryDir, "verify-result.json")
     verifyRequest.writeVerifyJson(groupedRules, verifyResponseFile)
-    exec.javaexec { e ->
-        e.mainClass.set("com.intellij.rt.coverage.verify.Main")
-        e.classpath = classpath
-        e.args = mutableListOf(verifyRequest.canonicalPath)
+    exec.javaexec {
+        mainClass.set("com.intellij.rt.coverage.verify.Main")
+        this@javaexec.classpath = classpath
```
aSemy, line 41: `kotlin-dsl` plugin makes the javaexec the receiver, thus necessitating `this@javaexec`. It's not very pretty - adjusting the arg name would help

src/main/kotlin/kotlinx/kover/appliers/KoverProjectApplier.kt
```diff
@@ -217,10 +218,10 @@ private fun Project.binaryReports(extension: KoverProjectConfig): List<File> {
...
 
-    return tasks.withType(Test::class.java).asSequence()
+    return tasks.withType<Test>().asSequence()
         // task can be disabled in the project extension
         .filterNot { extension.instrumentation.excludeTasks.contains(it.name) }
-        .map { t -> t.extensions.getByType(KoverTaskExtension::class.java) }
+        .map { t -> t.extensions.getByType<KoverTaskExtension>() }
         // process binary report only from tasks with enabled cover
         .filterNot { e -> e.isDisabled.get() }
         .map { e -> e.reportFile.get().asFile }
```
aSemy, line 227: this sort of code is a good candidate for replacing with a Gradle 'Configuration' that collects the outputs of tasks, to improve compatibility with the config cache. At the moment `asSequence()` will make the tasks run immediately, rather than on request.

src/main/kotlin/kotlinx/kover/appliers/KoverMergedApplier.kt
```diff
@@ -16,30 +16,36 @@ import kotlinx.kover.api.KoverPaths.MERGED_HTML_REPORT_DEFAULT_PATH
...
+        isEnabled.convention(false)
+        filters.classes.convention(KoverClassFilter())
+        filters.projects.convention(KoverProjectsFilter())
+        xmlReport.onCheck.convention(false)
+        xmlReport.reportFile.convention(layout.buildDirectory.file(MERGED_XML_REPORT_DEFAULT_PATH))
+        xmlReport.classFilter.convention(filters.classes)
+        htmlReport.onCheck.convention(false)
+        htmlReport.reportDir.convention(layout.buildDirectory.dir(MERGED_HTML_REPORT_DEFAULT_PATH))
+        htmlReport.classFilter.convention(filters.classes)
+        verify.onCheck.convention(false)
```
aSemy, line 47: using `convention()` to set default values is more idiomatic than `set()`

src/main/kotlin/kotlinx/kover/api/KoverConfig.kt
```diff
@@ -240,11 +240,11 @@ public open class KoverProjectHtmlConfig @Inject constructor(private val objects
 
 
 public open class KoverMergedConfig @Inject constructor(objects: ObjectFactory) {
-    internal var isEnabled: Property<Boolean> = objects.property(Boolean::class.java)
-    internal val filters: KoverMergedFilters = objects.newInstance(KoverMergedFilters::class.java, objects)
-    internal val xmlReport: KoverMergedXmlConfig = objects.newInstance(KoverMergedXmlConfig::class.java, objects)
-    internal val htmlReport: KoverMergedHtmlConfig = objects.newInstance(KoverMergedHtmlConfig::class.java, objects)
-    internal val verify: KoverVerifyConfig = objects.newInstance(KoverVerifyConfig::class.java, objects)
```
aSemy, line 247: Because the instances being created here have constructors annotated with `@Inject`, there's no need to pass in `objects` manually.

settings.gradle.kts
```diff
@@ -1,9 +1 @@
 rootProject.name = "kover"
-
-pluginManagement {
-    val kotlinVersion: String by settings
-
-    plugins {
-        kotlin("jvm") version kotlinVersion
-    }
-}
```
aSemy, line 9: this is removed because the version is now provided by Gradle's `embeddedKotlinVersion`

src/main/kotlin/kotlinx/kover/appliers/KoverMergedApplier.kt
```diff
@@ -169,10 +175,13 @@ private inline fun Project.mergedFilesProvider(
...
-            .map { (project, extension) -> project.path to project.projectFiles(filterExtractor(extension), extension) }
-            .associate { it }
+    return extensionByProject.map { extension ->
+        extension.map { (project, extension) ->
+            project.path to project.projectFiles(
+                filterExtractor(extension),
+                extension
+            )
+        }.associate { it }
     }
```
aSemy, line 185: using `providerX.map {}` (rather than `provider{ providerX.get()... }` means that if the provider is associated with a task, then that association remains. It is of course easy to confuse with Kotlin's `Collection.map {}` function...

src/main/kotlin/kotlinx/kover/appliers/KoverProjectApplier.kt
shanshin: The `asSequence()` itself does not launch tasks, and the call of this function is assumed only inside the provider. All this construction does not lead to the launch of tasks, other providers are designed for this.

src/test/kotlin/KoverPluginTest.kt
shanshin: Since this test does not test the functionality of the plugin and is used for local debugging, I think it is better to leave it in the local branch.
In addition, it slightly smears the testing levels across two modules - in this project there are  partial unit-tests, and not for the entire plugin, the dependency in it on `gradleTestKit()` is redundant.

In addition, remember that the functional tests run by Gradle itself are not entirely honest, because there are minor differences in the environment and dependencies. Therefore, for verification, I recommend publishing to the local repository `gradlew publishToMavenLocal` and already from `mavenLocal()` use the plugin.

settings.gradle.kts
shanshin: I think it's better to return the previous version, because Gradle uses old versions.
And if some bug is found in the compiler (for example  https://github.com/Kotlin/kotlinx-kover/blob/a0ff8d9c2c369dd1dbfd43f88e4c0169c262553d/src/main/kotlin/kotlinx/kover/appliers/KoverMergedApplier.kt#L92), then we will have to wait a long time for the release of Gradle, which will use the expected Kotlin.

aSemy: I've reverted it but it does lead to this warning from Gradle

```
> Configure project :
WARNING: Unsupported Kotlin plugin version.
The `embedded-kotlin` and `kotlin-dsl` plugins rely on features of Kotlin `1.5.31` that might work differently than in the requested version `1.7.10`.
```

I've never seen any problems come out of this warning, so I think it can be ignored.

I lean more towards using the same version as Gradle, because at least the compiler bug are known problems. It's never been clear to me what the problems are if the project is compiled with a different version (even if the Kotlin language level is 1.4).

src/test/kotlin/KoverPluginTest.kt
aSemy: It tests whether the Kover plugin can be applied, that's functionality :)

> Therefore, for verification, I recommend publishing to the local repository `gradlew publishToMavenLocal` and already from `mavenLocal()` use the plugin.

That's a lot more work than running a single unit test!

I found this test really useful when trying to track down why the other tests weren't working. The existing tests didn't give enough detail. I did try adjusting them, but I didn't really understand how they're set up and created. I find it much easier to see the build and settings files under test, and adjust the command and output log.

src/main/kotlin/kotlinx/kover/appliers/KoverProjectApplier.kt
aSemy: Sorry I meant 'be configured' instead of run. And yeah, okay, I see now that this function is only called from within a provider.

Anyway, it's not relevant to this PR, this can be revisited later.

src/test/kotlin/KoverPluginTest.kt
shanshin: Each of the available functional tests checks that the plugin is being applied, a separate test for this is useless.

In any case, example project launches will be added soon (see #219), so there is no need for a separate such test, especially in another module.

shanshin: The best way is to improve the error messages of the existing functional tests (adding the error log and the contents of the build script to the message, I will do it after merge #219), rather than adding a completely duplicate test.

shanshin (approved):
Great job, thanks!

PR #224, opened 2022-08-27
filter out projects without a build file 

aSemy:
A very quick fix for #222


src/functionalTest/kotlin/kotlinx/kover/test/functional/cases/MultiProjectTests.kt
```diff
@@ -190,4 +192,120 @@ internal class MultiProjectTests : BaseGradleScriptTest() {
             }
         }
     }
+
+    @Test
+    fun testNestedProjectInsideEmptyProject() {
```
shanshin, line 197: At the moment, it is good, but in the future it is necessary to create folder with examples of such special case tests.
It is worth separating the various tests, and the tests of special cases when there is no need to run the test in different languages and engines.

shanshin (approved):
Thanks!

PR #227, opened 2022-08-28
define test report extension in the report-vendor enum

aSemy:
Minor tweak that seems a little more clear to me :)


shanshin:
Thanks for contribution!

PR #230, opened 2022-09-01
Contribution guide

shanshin:


CONTRIBUTING.md
```diff
@@ -0,0 +1,77 @@
...
+  * If there is an alternative way to do what you need, then show the code of the alternative.
+
+## Submitting PRs
+
+We love PRs. Submit PRs [here](https://github.com/Kotlin/kotlinx-kover/pulls).
+However, please keep in mind that maintainers will have to support the resulting code of the project,
+so do familiarize yourself with the following guidelines. 
+
+* All development (both new features and bug fixes) is performed in the `main` branch.
+  * The `release` branch always contains sources of the most recently released version.
```
qwwdfsad, line 33: We don't have such a branch, do we?

qwwdfsad (approved):
Otherwise good

CONTRIBUTING.md
shanshin: It doesn't exist at the moment, but in this guide I suggested that such a branch could just so that there would be a way to open the code of the most recent release in addition to searching tags.

qwwdfsad: Good

PR #240, opened 2022-09-20
Implemented filtering of reports by annotation

shanshin:
* added annotation filtering API
* added functional tests
* XML and HTML report generation moved to Kover Aggregator
* default and minimal IntelliJ (Kover) Engine upgraded to `1.0.682`

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ kover {
             excludes += listOf("com.example.subpackage.*") // class exclusion rules
         }
         annotations { // common annotation filter for all default Kover tasks in this project
-            excludes += listOf("com.example.Annotation", "*Generated") // exclude classes and functions marked by specified annotations
+            excludes += listOf("com.example.Annotation", "*Generated") // exclude declarations marked by specified annotations
         }
     }
 
@@ -259,7 +259,7 @@ kover {
                 includes += "com.example2.*" // override class inclusion rules
                 excludes += listOf("com.example2.subpackage.*") // override class exclusion rules
             }
-            annotations { // override common annotation filter for XML report
+            annotations { // override common annotation filter for XML report (filtering will take place only by the annotations specified here)
                 excludes += listOf("com.example2.Annotation")
             }
         }
@@ -273,7 +273,7 @@ kover {
                 includes += "com.example2.*" // class inclusion rules
                 excludes += listOf("com.example2.subpackage.*") // override class exclusion rules
             }
-            annotations { // override common annotation filter for HTML report
+            annotations { // override common annotation filter for HTML report (filtering will take place only by the annotations specified here)
                 excludes += listOf("com.example2.Annotation")
             }
         }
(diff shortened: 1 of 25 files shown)


README.md
```diff
@@ -238,10 +238,13 @@ kover {
     isDisabled.set(false) // true to disable instrumentation and all Kover tasks in this project
     engine.set(DefaultIntellijEngine) // to change engine, use kotlinx.kover.api.IntellijEngine("xxx") or kotlinx.kover.api.JacocoEngine("xxx")
     filters { // common filters for all default Kover tasks
-        classes { // common class filter for all default Kover tasks 
+        classes { // common class filter for all default Kover tasks in this project
             includes += "com.example.*" // class inclusion rules
             excludes += listOf("com.example.subpackage.*") // class exclusion rules
         }
+        annotations { // common annotation filter for all default Kover tasks in this project
+            excludes += listOf("com.example.Annotation", "*Generated") // exclude classes and functions marked by specified annotations
```
qwwdfsad, line 246: Here and everywhere: "classes and functions" -> "declarations"

qwwdfsad, line 246: I also wonder if we have to treat `Generated` as excluded by default, with the only option to override it via `includes`? WDYT? How JaCoCo works here?

src/main/kotlin/kotlinx/kover/api/KoverConfig.kt
```diff
@@ -296,6 +312,15 @@ public open class KoverMergedFilters @Inject constructor(private val objects: Ob
         classes.set(classFilter)
     }
 
+    /**
+     * Configures annotation filter.
```
qwwdfsad, line 316: Please use consistent approach to the documentation and thoroughly document each public declaration: what are these filters and what for. Like you've done for `annotations` block above

qwwdfsad (commented):
Public API shape works for me modulo comments.

The left of the review, including the implementation I leave to @sandwwraith

src/main/kotlin/kotlinx/kover/api/KoverConfig.kt
sandwwraith, line 316: You can copy rich KDoc from `annotation` function higher in the file

src/main/kotlin/kotlinx/kover/api/KoverConfig.kt
```diff
@@ -328,10 +354,20 @@ public open class KoverMergedXmlConfig @Inject constructor(private val objects:
         config.execute(newClassFilter)
         classFilter.set(newClassFilter)
     }
+
+    /**
+     * Override annotation filter for the merged XML report generation task.
```
sandwwraith, line 359: I think documentation for 'override' action needs to be more specific here and in the example buildfile. It's not clear what 'override' does — does it remove previous filters? Or just adds new ones and select them in case of conflict?

README.md
shanshin: JaCoCo always filters only by the `"*Generated*"` annotation.
I propose at the first stage to let the user specify all annotations with which declarations are marked that they would like to exclude.

If there will be a high request from users that the `"*Generated*"` annotation should be used by default, then we can add a flag to add the `"*Generated*"` to the list (and by default it will be true).

sandwwraith (approved):

Issue #248, opened 2022-10-04
Kover 0.6.1: No instrumentation registered! Must run under a registering instrumentation.


soywiz:
After running `./gradlew koverHtmlReport` in a kotlin multiplatform project configuring JVM, android and K/N targets I get this exception:

https://github.com/korlibs/korge/actions/runs/3180418274/jobs/5184004872

```
Caused by:
java.lang.IllegalStateException: No instrumentation registered! Must run under a registering instrumentation.
   at androidx.test.platform.app.InstrumentationRegistry.getInstrumentation(InstrumentationRegistry.java:45)
   at androidx.test.core.app.ApplicationProvider.getApplicationContext(ApplicationProvider.java:41)
```

I suspect it is due to the android target. Is it possible to disable some targets just for the kover build? Is there any way to prevent that exception from happening?

soywiz:
Looking at the task it says `testDebugUnitTest` so I guess in the local machine as an Android unit test.
The thing is that im not running those Android unit tests under any circumstances. Im running connected tests in another CI job since I need the Application Context in some cases.
So I would like to prevent kover from using the Android unittests while still allowing to use connected tests when calling the original task directly

```
> Task :korau:testDebugUnitTest
[667](https://github.com/korlibs/korge/actions/runs/3180418274/jobs/5184004872#step:8:668)

[668](https://github.com/korlibs/korge/actions/runs/3180418274/jobs/5184004872#step:8:669)
com.soywiz.korau.format.AudioFormatTest > wavCorrupted FAILED
[669](https://github.com/korlibs/korge/actions/runs/3180418274/jobs/5184004872#step:8:670)
    java.lang.reflect.InvocationTargetException
[670](https://github.com/korlibs/korge/actions/runs/3180418274/jobs/5184004872#step:8:671)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[671](https://github.com/korlibs/korge/actions/runs/3180418274/jobs/5184004872#step:8:672)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[672](https://github.com/korlibs/korge/actions/runs/3180418274/jobs/5184004872#step:8:673)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[673](https://github.com/korlibs/korge/actions/runs/3180418274/jobs/5184004872#step:8:674)
[...]

PR #249, opened 2022-10-04
Adds the error message one would get when instrumenation is not enabled

soywiz:
As described here: https://github.com/Kotlin/kotlinx-kover/issues/248

So people can find the issue easier


shanshin (approved):

shanshin:
Thanks!

PR #253, opened 2022-10-11
Renamed Coverage Engine to Coverage Tool and IntelliJ Engine to Kover Tool

shanshin:


CONTRIBUTING.md
```diff
@@ -11,8 +11,8 @@ Submit issues [here](https://github.com/Kotlin/kotlinx-kover/issues).
 * Search for existing issues to avoid reporting duplicates.
 * When submitting a bug report:
   * Test it against the most recently released version. It might have been already fixed.
-  * By default, we assume that your problem reproduces in Kotlin/JVM project with using the IntelliJ Coverage Engine. Please, mention if the problem is
-    specific to Kotlin Multi-Platform/Android or JaCoCo. 
+  * By default, we assume that your problem reproduces in Kotlin/JVM project with using the Kover Coverage Tool. Please, mention if the problem is
```
qwwdfsad, line 14:
```suggestion
  * By default, we assume that your problem reproduces in Kotlin/JVM project with the Kover Coverage Tool. Please, mention if the problem is
```

CONTRIBUTING.md
```diff
@@ -11,8 +11,8 @@ Submit issues [here](https://github.com/Kotlin/kotlinx-kover/issues).
...
+    specific to Kotlin Multi-Platform/Android or JaCoCo Tool.
```
qwwdfsad, line 15:
```suggestion
    specific to Kotlin Multiplatform/Android or JaCoCo Tool. 
```

README.md
```diff
@@ -111,7 +111,7 @@ Merged reports are reports that combine statistics of code coverage by test task
 At the same time, for each project, its configuration of instrumentation and special filters (non-class filters) is applied.
 
 For every Gradle project participating in the merged report, both Kover plugin should be applied
-and Coverage Engine types and versions should be identical.
+and Coverage Tool types and versions should be identical.
```
qwwdfsad, line 114: Could you please elaborate on the choice "Engine -> Tool"? 
Engine seems far more natural for me, but I might be mistaken

README.md
```diff
@@ -237,7 +237,7 @@ In the project in which the plugin is applied, you can configure instrumentation
 ```kotlin
 kover {
     isDisabled.set(false) // true to disable instrumentation and all Kover tasks in this project
-    engine.set(DefaultIntellijEngine) // to change engine, use kotlinx.kover.api.IntellijEngine("xxx") or kotlinx.kover.api.JacocoEngine("xxx")
+    tool.set(DefaultKoverTool) // to change tool, use kotlinx.kover.api.KoverTool("xxx") or kotlinx.kover.api.JacocoTool("xxx")
```
qwwdfsad, line 240:
```suggestion
    tool.set(DefaultKoverTool) // to change the tool, use kotlinx.kover.api.KoverTool("xxx") or kotlinx.kover.api.JacocoTool("xxx")
```

README.md
shanshin: In fact, the agent and other libraries are something third-party to the Kover Gradle Plugin (KGP). 
If we perceive them as something external, third-party, something auxiliary, then it began to seem to me that the word "engine" is more associated with something internal, its own code. In addition, the plugin itself may have some kind of its own "engine" entity inside.

When I was trying to find an alternative, I noticed that JaCoCo is most often called a tool (or toolkit), so I thought it was quite an appropriate name (e.g. https://www.eclemma.org/ , in [teamcity]( https://www.jetbrains.com/help/teamcity/jacoco.html) or `toolVersion` in [Gradle](https://docs.gradle.org/current/userguide/jacoco_plugin.html#sec:configuring_the_jacoco_plugin) ).

In addition, if we add a CLI over the agent and start delivering it separately without KGP, then "Kover Coverage Tool" is a more telling name than "Kover Coverage Engine".

qwwdfsad: Thanks, this reasoning works for me

README.md
```diff
@@ -4,7 +4,7 @@
 [![JetBrains incubator project](https://jb.gg/badges/incubator.svg)](https://confluence.jetbrains.com/display/ALL/JetBrains+on+GitHub)
 [![GitHub license](https://img.shields.io/badge/license-Apache%20License%202.0-blue.svg?style=flat)](https://www.apache.org/licenses/LICENSE-2.0)
 
-**Kover** - Gradle plugin for Kotlin code coverage agents: [IntelliJ](https://github.com/JetBrains/intellij-coverage)
+**Kover** - Gradle plugin for Kotlin code coverage tools: [Kover](https://github.com/JetBrains/intellij-coverage)
```
sandwwraith, line 7: While I'm OK with Engine -> Tool renaming, I'm not sure people would understand the difference between Kover (Gradle plugin) and Kover (Coverage tool). It may confuse people when we try to explain to them they can use Kover with Jacoco Tool and Kover with Kover tool. Also, the repository name is still `intellij-coverage`. Have you considered just naming it `Intellij Tool`?

docs/migration-to-0.7.0.md
```diff
@@ -0,0 +1,76 @@
+# Kover migration guide from 0.6.x to 0.7.0
+
+## Main differences
+- coverage engines was renamed to coverage tools
```
sandwwraith, line 4:
```suggestion
- coverage engines were renamed to coverage tools
```

README.md
shanshin: The repository is also used for IDEA, and at the moment the benefits of renaming it are doubtful, and it will be quite problematic.

The mention of `IntelliJ` in the name of the tool is also a bit misleading, because it has no direct relation to the `IntelliJ platform`.

I see the main development prospect that the Carpet should become an independent project, independent of the way it is used. That is, whether it is a Gradle Plugin, Kotlin Compiler plugin or CLI or maven in the future, they all just have to use the Kover.

In this case, the main problem is that the plugin and the tool will be called the same. In JaCoCo, it is called similarly, but its plugin supports only one tooling.

Might be worth doing some extra design on that?

README.md
```diff
@@ -237,7 +237,7 @@ In the project in which the plugin is applied, you can configure instrumentation
 ```kotlin
 kover {
     isDisabled.set(false) // true to disable instrumentation and all Kover tasks in this project
-    engine.set(DefaultIntellijEngine) // to change engine, use kotlinx.kover.api.IntellijEngine("xxx") or kotlinx.kover.api.JacocoEngine("xxx")
+    tool.set(DefaultKoverTool) // to change the tool, use kotlinx.kover.api.KoverTool("xxx") or kotlinx.kover.api.JacocoTool("xxx")
```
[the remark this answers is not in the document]
shanshin: M.b. `KoverToolDefault` (similar to `KoverTool("1.0.654")`)?

IgnatBeresnev (approved):
I was asked to proofread documentation changes. I don't see any major mistakes worth correcting that were introduced in this particular PR, but I do see some in other parts of the `README.md` file: some are objective, some are purely subjective and stylistic

So instead of arguing arguing about it here and dragging out review, I'll submit another PR with my corrections. I'll wait for this one to be merged first though to avoid merge conflicts on my end

qwwdfsad (approved):
No more comments from me apart from Leonid's ones

README.md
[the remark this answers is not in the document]
sandwwraith: Yeah, that would be better

sandwwraith (approved):

Issue #111, opened 2021-12-29
Kover and Gradle tests with Gradle Test Kit


JavierSegoviaCordoba:
I have a multiproject project where there are a lot of Gradle Plugins with tests that are using `gradleTestKit()`. If I add Kover to the project, Gradle tests that use Gradle Test Kit start to fail.

Repro: https://github.com/JavierSegoviaCordoba/gradle-plugins/tree/kover-repro

Comment/uncomment Kover in the root repo plugins block to reproduce the issue.

Looks like Kover is blocking my plugins can be added to the test classpath by Gradle Test Kit.

JavierSegoviaCordoba:
Tried with 0.5.0-RC2 but no luck, it still fails.

shanshin:
Hi,
when applying the plugin, an identical path is used. 
The use of the plugin itself does not affect the functional tests in any way, you can make sure of this if you uncomment the `javiersc-code-coverage` in the root project, and add 
```
tasks.test {
    extensions.configure(kotlinx.kover.api.KoverTaskExtension::class) {
        isDisabled = true
    }
}
```
 to some project - then functional tests will work in this project.

For some reason, using an javaagent to instrumentation functional tests causes an error. You can check this by applying the `jacoco` plugin to any project - functional tests will fall.

Further study is needed to understand why any code instrumentation breaks functional tests.

JavierSegoviaCordoba:
This should work on 0.7.0, right?

PR #262, opened 2022-10-25
Redesigned the launch of functional tests for the use of system processes

shanshin:
Using the Gradle test toolkit runs functional tests with the wrong environment that should be in a real project: dependencies that have not been declared are added (the `kotlinx.kover.test.functional.cases.AdaptersTests` was always executed successfully, because it is always K/MP plugin in classpath); using the plugin causes an [error](https://github.com/Kotlin/kotlinx-kover/issues/111); Android projects are not supported.

As a solution to these problems, the use of `GradleRunner` has been replaced by running Gradle system processes (using system shell, like `sh` or `cmd`).


qwwdfsad:
Please rebase on `develop` and fix build issue that makes build failing

docs/development/functionalTests.md
```diff
@@ -0,0 +1,25 @@
+# Functional Tests
+Functional tests run Gradle in a separate system process, and then check the result of the work.
+
+By default, to run functional tests, Gradle is used, which builds a plugin.
```
qwwdfsad, line 4:
```suggestion
By default, to run functional tests, Gradle is used, which builds the plugin as its dependency.
```

examples/android_kts/build.gradle.kts
```diff
@@ -6,12 +6,12 @@ buildscript {
     dependencies {
         classpath("com.android.tools.build:gradle:${Versions.ANDROID_GRADLE_PLUGIN}")
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:${Versions.KOTLIN}")
-        classpath("org.jetbrains.kotlinx:kover:${Versions.KOVER}")
+        classpath("org.jetbrains.kotlinx:kover:0.6.1")
```
qwwdfsad, line 9: The key here is not to forget update this version once in a while, not sure how we are going to ensure it

qwwdfsad (commented):
LGTM in general.

Prior to merging, please update the PR description (and later copy it to the squashed commit message) that explains **why** this change is done and what problem it solves.

Otherwise nobody will able to recall why exactly we changed the say tests are ran and what for

qwwdfsad (approved):

PR #266, opened 2022-10-27
Revise documentation

IgnatBeresnev:
As promised, I went over documentation and here are my proposals. Sorry, got carried away a little bit.

DISCLAIMER: I'm neither a technical writer nor a native English speaker, but I do have some experience with English and reading good and bad documentation :) So take everything with a grain of salt, feel free to revert any changes you don't agree with and ask questions.

More importantly, I'm not a Kover expect by far, so please proofread everything :) I might've misunderstood some things.

Changes made throughout the whole file:

* Quickstart now truly offers a quick start - it was strange to see the tasks all the way in the back. As a developer, I just want to copy-paste the plugin and see which task I need to run. Done. If there's any problems, I'll come back and read the docs.
* Made all lines be <= 120 chars. It's better for review (no side scrolling) and for reading and editing the file in the general.
* Simplified sentences a little bit, changed some passive voice to active. Pretty much applied some suggestions from [Kotlin documentation guidelines](https://docs.google.com/document/d/1mUuxK4xwzs3jtDGoJ5_zwYLaSEl13g_SuhODdFuh2Dc/edit#heading=h.hjojzyai18a6). Not everything is perfect, but I tried. 
* Added some spacing before and after elements, including headers, code blocks and spoilers. Some tools that render markdown files don't work well if there are no newlines before and after such elements. While it didn't seem to make any difference for GitHub, I think it's just a good thing to have.

I'll try to additionally explain some specific changes as review comments

Here's a link to preview changes just in case: https://github.com/Kotlin/kotlinx-kover/blob/revise-readme/README.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ Submit issues [here](https://github.com/Kotlin/kotlinx-kover/issues).
 * Search for existing issues to avoid reporting duplicates.
 * When submitting a bug report:
   * Test it against the most recently released version. It might have been already fixed.
-  * By default, we assume that your problem can be reproduced in a Kotlin/JVM project with the Kover Coverage Tool.
+  * By default, we assume that your problem is reproduced in a Kotlin/JVM project with the Kover Coverage Tool.
     Please, mention if the problem is specific to Kotlin Multiplatform/Android or JaCoCo Tool. 
   * Include the code that reproduces the problem. Provide the complete reproducer code, yet minimize it as much as possible.
   * However, don't put off reporting any weird or rarely appearing issues just because you cannot consistently 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,8 +11,8 @@ Submit issues [here](https://github.com/Kotlin/kotlinx-kover/issues).
 * Search for existing issues to avoid reporting duplicates.
 * When submitting a bug report:
   * Test it against the most recently released version. It might have been already fixed.
-  * By default, we assume that your problem reproduces in Kotlin/JVM project with the Kover Coverage Tool. Please, mention if the problem is
-    specific to Kotlin Multiplatform/Android or JaCoCo Tool. 
+  * By default, we assume that your problem can be reproduced in a Kotlin/JVM project with the Kover Coverage Tool.
+    Please, mention if the problem is specific to Kotlin Multiplatform/Android or JaCoCo Tool. 
   * Include the code that reproduces the problem. Provide the complete reproducer code, yet minimize it as much as possible.
   * However, don't put off reporting any weird or rarely appearing issues just because you cannot consistently 
     reproduce them.
--- a/README.md
+++ b/README.md
@@ -314,7 +314,7 @@ You can configure Kover, its tasks and instrumentation in any project for which 
 <summary>Kotlin</summary>
 
 ```kotlin
-kover { 
+kover {
     isDisabled.set(false) // true to disable instrumentation and all Kover tasks in this project
     tool.set(KoverToolDefault) // to change the tool, use kotlinx.kover.api.KoverTool("xxx") or kotlinx.kover.api.JacocoTool("xxx")
     filters { // common filters for all default Kover tasks
--- a/README.md
+++ b/README.md
@@ -636,6 +636,7 @@ You can choose which Coverage Tool and which version to use when configuring Kov
 ```kotlin
 kotlinx.kover.api.KoverToolDefault
 ```
+
 </details>
 
 <details>
@@ -644,20 +645,24 @@ kotlinx.kover.api.KoverToolDefault
 ```groovy
 kotlinx.kover.api.KoverToolDefault.INSTANCE
 ```
+
 </details>
 
 #### Kover Coverage Tool with custom version
+
 ```
 kotlinx.kover.api.KoverTool("1.0.683")
 ```
 
 #### JaCoCo Coverage Tool with default version
+
 <details open>
 <summary>Kotlin</summary>
 
 ```kotlin
 kotlinx.kover.api.JacocoToolDefault
 ```
+
 </details>
 
 <details>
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-# Kotlinx-Kover
+# kotlinx-kover
```
IgnatBeresnev, line 1: Looked at other kotlinx libraries and it seems like [datetime](https://github.com/Kotlin/kotlinx-datetime), [lincheck](https://github.com/Kotlin/kotlinx-lincheck), [reflect lite](https://github.com/Kotlin/kotlinx.reflect.lite), [cli](https://github.com/Kotlin/kotlinx-cli) and [html](https://github.com/Kotlin/kotlinx.html) all stick to this lower-case artifact name-like style, so changed it for consistency

Alternatively, it could be changed to just `Kover`, it's the `kotlinx` part that's triggering me a bit

README.md
```diff
@@ -7,38 +7,38 @@
 **Kover** - Gradle plugin for Kotlin code coverage tools: [Kover](https://github.com/JetBrains/intellij-coverage)
 and [JaCoCo](https://github.com/jacoco/jacoco).
 
-Minimal supported `Gradle` version: `6.6`.
+Minimum supported version of `Gradle` is `6.6`.
```
IgnatBeresnev, line 10: Couldn't find any credible sources for this one, but googling for "Minimum version" yield way more results than vs "Minimal version", so `Minimum` _seems_ to be the correct word in this context. But both sound kinda weird to my ear tbh

README.md
```diff
@@ -7,38 +7,38 @@
...
 
 ## Table of contents
+
 - [Features](#features)
 - [Quickstart](#quickstart)
-  - [Apply plugin to a project](#apply-plugin)
-    - [Applying plugins with the plugins DSL](#applying-plugins-with-the-plugins-dsl)
-    - [Legacy Plugin Application: applying plugins with the buildscript block](#legacy-plugin-application-applying-plugins-with-the-buildscript-block)
```
IgnatBeresnev, line 17: It's not that far away from ToC, so probably no point in it taking 3 lines

README.md
```diff
@@ -7,38 +7,38 @@
...
+* Verification rules with bounds to keep track of coverage.
+* Customizable filters for instrumented classes.
 
 ## Quickstart
-### Apply plugin
-#### Applying plugins with the plugins DSL
-In top-level build file:
+
+The recommended way of applying Kover is with the 
+[plugins DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block).
```
IgnatBeresnev, line 39: Decided to stress it since the legacy method has some restrictions when it comes to type-safe DSL

README.md
```diff
@@ -150,18 +189,29 @@ tasks.test {
...
+        disabled = false
+      
+        // set file name of binary report
+        binaryReportFile.set(file("$buildDir/custom/result.bin"))
+      
+        // for details, see "Instrumentation inclusion rules" below
+        includes = ['com.example.*']
+      
+        // for details, see "Instrumentation exclusion rules" below
+        excludes = ['com.example.subpackage.*']
```
IgnatBeresnev, line 203: Here and in similar places it's just being nice to the reader so that they don't have to side scroll. 

I didn't touch any of the big blocks with settings as it's taking quite a bit of time, but I would recommend to do the same thing there. Just open the main configuration section and try to read some longer descriptions in the beginning - you'll need to scroll right, but the scroll bar is hidden below and so you need to scroll down first, then scroll right, then scroll up again and only then you can read the description.

README.md
```diff
@@ -709,9 +789,12 @@ An example is available [here](examples/android_groovy)
 </details>
 
 ## Implicit plugin dependencies
-While the plugin is being applied, the artifacts of the JaCoCo or Kover toolkit are dynamically loaded. They are downloaded from the `mavenCentral` repository.
 
-For the plugin to work correctly, you need to make sure that the `mavenCentral` (or any of its mirrors) is added to the repository list of the project in which the plugin is applied, if it doesn't already exist (usually this is the root project):
+While the plugin is being applied, the artifacts of the JaCoCo or Kover toolkit are dynamically loaded. They are
```
IgnatBeresnev, line 793: I don't quite understand what `While the plugin is being applied` means - do you mean when you've just added it or when you're running Kover? It sounds like the first, but then it seems kinda obvious that Kover would download dependencies and fail if it can't find an artifact?

README.md
shanshin: Perhaps it is better to write `kover gradle plugin`, because in fact this repository contains only a plugin for the Gradle build system, in the future it is possible to create a repository for Maven as well.

README.md
shanshin: This is OK, because I plan that only the applying of the plugin will remain in the README, and all the rest of the content will be moved to the HTML documentation of the release.

README.md
shanshin: This is approximately both.
[Applying plugin](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block) is a code like  
```
plugins {
    id("...")
}
```

The plugin is applied every time the build is run, even if no Kover command is explicitly invoked.
However, it is difficult to say unequivocally when exactly the dependencies are resolved - directly at the time of applying plug-in, or at the start of the Kover commands.

Perhaps it is better to reformulate: "during the running of the build ..."

CONTRIBUTING.md
```diff
@@ -11,8 +11,8 @@ Submit issues [here](https://github.com/Kotlin/kotlinx-kover/issues).
 * Search for existing issues to avoid reporting duplicates.
 * When submitting a bug report:
   * Test it against the most recently released version. It might have been already fixed.
-  * By default, we assume that your problem reproduces in Kotlin/JVM project with the Kover Coverage Tool. Please, mention if the problem is
```
shanshin, line 14: Initially, the wording was not quite correct. What was meant here: unless otherwise stated, it is assumed that the Kotlin/JVM plugin and Kover tool were used in the project.

shanshin (approved):

PR #281, opened 2022-11-22
Added Gradle minimum version check

shanshin:
At the moment, when trying to use a plugin with a Gradle version less than the minimum, the user receives non-obvious errors.
In order to explicitly tell the user the reason, we need to add a check at the plugin.


qwwdfsad (approved):
👍

PR #307, opened 2023-02-23
Adding binary compatibility validator

### LINKED ISSUE #305: Incorporate binary-compatiblity-validator
Starting from #295, Kover public API is becoming more and more stable, as well as (reasonably) stretched between implementation classes.

It would be nice to start using binary compatibility validator in order to ensure that we neither accidentally break API compatibility nor leak internal classes into public API

Kantis:
Fixes #305 
Awaiting merge of #295

### DIFF
--- a/api/kover.api
+++ b/api/kover.api
@@ -1,0 +1,268 @@
+public final class kotlinx/kover/KoverPlugin : org/gradle/api/Plugin {
+	public fun <init> ()V
+	public synthetic fun apply (Ljava/lang/Object;)V
+	public fun apply (Lorg/gradle/api/Project;)V
+}
+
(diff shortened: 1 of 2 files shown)


shanshin:
Hi, thanks for the effort!

> Should perhaps await merge of https://github.com/Kotlin/kotlinx-kover/pull/295 ?

Yes, the API will change significantly there.

qwwdfsad:
>add apiCheck task to TeamCity pipeline

There is no need to, BCV will do it automatically

shanshin:
Thanks for the effort, however, after reviewing the API dump, it became clear that it was necessary to rework the API a little. 
So this PR superseded by #311

PR #311, opened 2023-03-01
Added binary compatibility validator

### LINKED ISSUE #305: Incorporate binary-compatiblity-validator

shanshin:
Resolves #305


qwwdfsad (approved):
The next step is to ensure each entity in the ABI has its own documentation

shanshin:
> The next step is to ensure each entity in the ABI has its own documentation

#323

PR #328, opened 2023-03-27
Added dokka docs for DSL

shanshin:


build.gradle.kts
```diff
@@ -115,6 +115,17 @@ tasks.withType<KotlinCompile>().configureEach {
     }
 }
 
+tasks.dokkaHtml {
```
qwwdfsad, line 118: What puzzles me -- who (and when) is responsible for updating the doc and the site?

qwwdfsad (approved):

PR #383, opened 2023-05-25
Updated Gradle and Kotlin versions

shanshin:
Gradle version upgraded to 8.1.1 - appeared new messages in the configuration cache report, the `capitalized` utility method disappeared
Kotlin version upgraded to 1.8.20 - Android tests stopped working, a separate public class for SAM conversion appeared


src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/VerificationTests.kt
```diff
@@ -125,20 +125,20 @@ Rule violated: lines covered percentage is 46.5900, but expected minimum is 58.0
 Rule violated: branches covered count for class 'org.jetbrains.kover.test.functional.verification.FullyCovered' is 0, but expected minimum is 1000
 Rule violated: instructions missed percentage for class 'org.jetbrains.kover.test.functional.verification.FullyCovered' is 0.0000, but expected minimum is 100.0000
 Rule violated: branches covered count for class 'org.jetbrains.kover.test.functional.verification.PartiallyCoveredFirst' is 2, but expected minimum is 1000
-Rule violated: instructions missed percentage for class 'org.jetbrains.kover.test.functional.verification.PartiallyCoveredFirst' is 43.3300, but expected minimum is 100.0000
+Rule violated: instructions missed percentage for class 'org.jetbrains.kover.test.functional.verification.PartiallyCoveredFirst' is *, but expected minimum is 100.0000
```
sandwwraith, line 128: I'm not sure this is a good change, because desired percentage is an important information

shanshin: These values relate only to the instruction, they change every time the compiler version is updated.

Therefore, either you need to change the text every such time, or figure out how to generalize it. Using `*` is the easiest option.

sandwwraith: So in real life, there would be an actual number? Ok then

sandwwraith (approved):

PR #395, opened 2023-06-01
Added a filter of source sets

### LINKED ISSUE #376: How to exclude test fixtures?
I have a project that uses the `java-test-fixtures` plugin. When I run the `koverHtmlReport` tasks, classes in the `testFixtures` source set are included. I see that there are old issues on this, but none have a solution in them.

I've tried this
```kts
kover {
    excludeTests {
        tasks("testFixturesClasses")
    }
}
```
But they are still included in the report. Any help would be appreciated.

shanshin:
Resolves #376

### DIFF
--- a/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverProjectExtension.kt
+++ b/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverProjectExtension.kt
@@ -74,7 +74,20 @@ public interface KoverProjectExtension {
      */
     public fun excludeInstrumentation(config: Action<KoverInstrumentationExclusions>)
 
-
+    /**
+     * Excludes classes of the specified source sets from Kover reports.
+     *
+     * As a side effect, the generation of Kover reports ceases to depend on the compilation tasks of these source sets.
+     *
+     * Example:
+     * ```
+     * kover {
+     *     excludeSourceSets {
+     *         names("test1", "extra")
+     *     }
+     * }
+     * ```
+     */
     public fun excludeSourceSets(config: Action<SourceSetsExclusions>)
 
     /*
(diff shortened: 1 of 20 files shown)


src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverProjectExtension.kt
```diff
@@ -74,6 +74,9 @@ public interface KoverProjectExtension {
      */
     public fun excludeInstrumentation(config: Action<KoverInstrumentationExclusions>)
 
+
+    public fun excludeSourceSets(config: Action<SourceSetsExclusions>)
```
sandwwraith, line 78: Documentation for function is missing

src/main/kotlin/kotlinx/kover/gradle/plugin/locators/Commons.kt
```diff
@@ -5,10 +5,13 @@
 package kotlinx.kover.gradle.plugin.locators
 
 import kotlinx.kover.gradle.plugin.commons.CompilationUnit
+import kotlinx.kover.gradle.plugin.dsl.internal.KoverProjectExtensionImpl
 import kotlinx.kover.gradle.plugin.util.DynamicBean
 import org.gradle.api.Task
 import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.tasks.SourceSet
 import org.gradle.api.tasks.TaskProvider
+import org.jetbrains.kotlin.gradle.utils.`is`
```
sandwwraith, line 14: I don't see these usages. Redundant import?

sandwwraith (approved):

src/main/kotlin/kotlinx/kover/gradle/plugin/locators/Commons.kt
shanshin: Yes

PR #403, opened 2023-06-02
Performed code refactoring

shanshin:
Due to the increase in the number of artifacts and dependencies, in order to avoid errors in the future, it became necessary to change the structure of subprojects and add a version catalog.

Gradle plugin is now a subproject, on a par with the CLI and others.

Also implemented is getting rid of buildSrc as an outdated approach in favor of composite assemblies.


gradle/libs.versions.toml
```diff
@@ -0,0 +1,25 @@
...
+
+[libraries]
+
+# IntelliJ coverage library
+intellij-reporter = { module = "org.jetbrains.intellij.deps:intellij-coverage-reporter", version.ref = "intellij-coverage" }
+intellij-offline = { module = "org.jetbrains.intellij.deps:intellij-coverage-offline", version.ref = "intellij-coverage" }
+
+# JUnit
+junit-jupiter = { module = "org.junit.jupiter:junit-jupiter", version.ref = "junit" }
+junit-params = { module = "org.junit.jupiter:junit-jupiter", version.ref = "junit" }
```
sandwwraith, line 17: Are these coordinates supposed to be the same as `junit-jupiter`?

kover-gradle-plugin/build.gradle.kts
```diff
@@ -40,20 +38,18 @@ kotlin.target.compilations.run {
...
 
-    compileOnly("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion")
-
-    compileOnly("org.jetbrains.intellij.deps:intellij-coverage-reporter:1.0.721")
-
-    testImplementation(kotlin("test"))
-
-    "functionalTestImplementation"("org.junit.jupiter:junit-jupiter:5.9.0")
-    "functionalTestImplementation"("org.junit.jupiter:junit-jupiter-params:5.9.0")
+    "functionalTestImplementation"(kotlin("test"))
```
sandwwraith, line 44: is it possible to use a separate variable for configuration instead of using `Stirng.invoke()`? I think it would be more readable

settings.gradle.kts
```diff
@@ -1,11 +1,19 @@
-rootProject.name = "kover-gradle-plugin"
+rootProject.name = "kover"
 
 pluginManagement {
-    val kotlinVersion: String by settings
+    includeBuild("build-logic")
+
     plugins {
-        kotlin("jvm") version kotlinVersion
+        kotlin("jvm") version embeddedKotlinVersion
```
sandwwraith, line 7: Is `embeddedKotlinVersion` a variable provided by Gradle itself? I didn't notice its definition

settings.gradle.kts
```diff
@@ -1,11 +1,19 @@
...
+dependencyResolutionManagement {
+    versionCatalogs {
+        create("libs")
     }
 }
 
-include(":toolset:kover-cli")
-include(":toolset:kover-offline-runtime")
\ No newline at end of file
+include(":kover-gradle-plugin")
```
sandwwraith, line 17: Also, I haven't noticed any changes in publication-related functionality. Are you sure root project won't be published accidentally?

sandwwraith (approved):

gradle/libs.versions.toml
shanshin: My bad)

settings.gradle.kts
shanshin: Yes, https://github.com/gradle/kotlin-dsl-samples/issues/94

settings.gradle.kts
shanshin: Yes, the root project cannot be published, because the `maven-publish` plugin is not applied in it and, accordingly, there are no publishing tasks.

PR #411, opened 2023-06-16
Added ability to generate reports even if there are no tests in the project

### LINKED ISSUE #409: Add the ability to generate reports even if there are no tests in the project
From the point of view of using the plugin, it is illogical when reports stop being generated if all tests are excluded or tests have not yet been added to the project.

With missing tests, XML/HTML reports should be generated with zero coverage, and verification should be carried out with the assumption that all lines/methods/classes are not covered.

shanshin:
Resolves #409

### DIFF
--- a/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/MultiProjectTests.kt
+++ b/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/MultiProjectTests.kt
@@ -114,6 +114,9 @@ internal class MultiProjectTests {
             subproject(subprojectPath) {
                 checkDefaultBinReport(false)
                 taskNotCalled(defaultTestTaskName(slice.type))
+                checkOutcome("koverXmlReport", "SUCCESS")
+                checkOutcome("koverHtmlReport", "SUCCESS")
+                checkOutcome("koverVerify", "SUCCESS")
             }
         }
     }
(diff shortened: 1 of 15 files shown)


kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/NoTestReportsTests.kt
```diff
@@ -0,0 +1,37 @@
...
+    private fun checkReportsGenerated(build: GradleBuild) {
+        val buildResult = build.runWithParams("koverXmlReport", "koverHtmlReport", "koverVerify")
+        val checkerContext = build.createCheckerContext(buildResult)
+
+        checkerContext.xml(defaultXmlReport()) {
+            classCounter("kotlinx.kover.templates.ExampleClass").assertFullyMissed()
+        }
+        checkerContext.checkHtmlReport()
+        assertFalse(buildResult.isSuccessful)
+        assertContains(buildResult.output, "Rule violated: lines covered percentage is 0.000000, but expected minimum is 50")
```
sandwwraith, line 34: Should there be a test without tests and without rules — to check that report with 0 coverage generated, but the build is still succesfull?

kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/MultiProjectTests.kt
```diff
@@ -113,16 +109,11 @@ internal class MultiProjectTests {
 
         run("koverXmlReport", "koverHtmlReport", "koverVerify") {
             checkDefaultBinReport(false)
-
-            checkOutcome("koverHtmlReport", "SKIPPED")
-            checkOutcome("koverXmlReport", "SKIPPED")
-            checkOutcome("koverVerify", "SKIPPED")
+            taskNotCalled(defaultTestTaskName(slice.type))
```
sandwwraith, line 112: Shouldn't we do `checkOutcome("koverHtmlReport", "SUCCESS")`?

sandwwraith (approved):

PR #412, opened 2023-06-16
Added DSL for Kover task names

shanshin:
With extension functions and val-s, it is allowed in kts scripts to write like `tasks.koverXmlReportName`.

Valid statements in DSL:
- `tasks.koverXmlReportName`
- `tasks.koverHtmlReportName`
- `tasks.koverVerifyName`
- `tasks.koverAndroidXmlReportName("variant")`
- `tasks.koverAndroidHtmlReportName("variant")`
- `tasks.koverAndroidVerifyName("variant")`
- `extensions.koverExtensionName`
- `extensions.koverReportExtensionName`


kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverNames.kt
```diff
@@ -0,0 +1,124 @@
...
+     */
+    public fun androidVerify(buildVariant: String): String {
+        return verifyTaskName(buildVariant)
+    }
+}
+
+/**
+ * Name of the XML report generation task for Kotlin JVM and Kotlin multiplatform projects.
+ *
+ * Same as a [KoverNames.DEFAULT_XML_REPORT_NAME].
```
sandwwraith, line 65: If they're the same, why `object KoverNames` should be exposed at all?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverNames.kt
```diff
@@ -0,0 +1,124 @@
+package kotlinx.kover.gradle.plugin.dsl
+
+import kotlinx.kover.gradle.plugin.commons.htmlReportTaskName
+import kotlinx.kover.gradle.plugin.commons.verifyTaskName
+import kotlinx.kover.gradle.plugin.commons.xmlReportTaskName
+import org.gradle.api.plugins.ExtensionContainer
+import org.gradle.api.tasks.TaskContainer
+
+public object KoverNames {
```
sandwwraith, line 9: Given that it is `public`, it is likely worth adding some documentation

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverNames.kt
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Name of the configurations to add dependency on Kover reports from another project.
```
sandwwraith, line 11:
```suggestion
     * Name of the configuration to add dependency on Kover reports from another project.
```

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverNames.kt
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * Same as a [KoverNames.DEFAULT_XML_REPORT_NAME].
+ */
+public val TaskContainer.koverXmlReportName
+    get() = KoverNames.DEFAULT_XML_REPORT_NAME
+
+/**
+ * Name of the HTML report generation task for Kotlin JVM and Kotlin multiplatform projects.
+ *
+ * Same as a [KoverNames.DEFAULT_HTML_REPORT_NAME].
```
sandwwraith, line 73:
```suggestion
 * Same as [KoverNames.DEFAULT_HTML_REPORT_NAME].
```

Here and below

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverNames.kt
shanshin: Then we will need `koverXmlReportName` to call, we will need instance of `TaskContainer`, but it may not be available in some places.

sandwwraith: Why extension on TaskContainer is needed then, if receiver is not used?

shanshin: For a more convenient search constants without studying the documentation. 
There is no need to know about the existence of the `KoverNames` object, just type `tasks.kover` and the IDE will offer these constants

sandwwraith (approved):

PR #424, opened 2023-07-06
Implemented task of generating binary report

### LINKED ISSUE #423: Make a path for the resulting .ic report configurable
In order to configure Qodana Code Coverage, we need to store ic report in the particular folder (probably having some particular name).

Being able to configure the output path like we have it for XML reports now.
Having a separate task to generate such a file.

shanshin:
Resolves #423

### DIFF
--- a/docs/gradle-plugin/configuring.md
+++ b/docs/gradle-plugin/configuring.md
@@ -344,6 +344,36 @@ koverReport {
             }
         }
 
+        // configure IC report for `release` build variant (task `koverIcReportRelease`)
+        ic {
+            //  generate an IC report when running the `check` task
+            onCheck.set(false)
+    
+            // IC report file
+            icFile.set(layout.buildDirectory.dir("my-project-report/report-release.ic"))
+    
+            // overriding filters only for the IC report
+            filters {
+                // exclusions for IC reports
+                excludes {
+                    // excludes class by fully-qualified JVM class name, wildcards '*' and '?' are available
+                    classes("com.example.*")
+                    // excludes all classes located in specified package and it subpackages, wildcards '*' and '?' are available
+                    packages("com.another.subpackage")
+                    // excludes all classes and functions, annotated by specified annotations (with BINARY or RUNTIME AnnotationRetention), wildcards '*' and '?' are available
+                    annotatedBy("*Generated*")
+                }
+      
+                // inclusions for IC report
+                includes {
+                    // includes class by fully-qualified JVM class name, wildcards '*' and '?' are available
+                    classes("com.example.*")
+                    // includes all classes located in specified package and it subpackages
+                    packages("com.another.subpackage")
+                }
+            }
+        }      
+      
         // configure verification for `release` build variant (task `koverVerifyRelease`)
         verify {
             //  verify coverage when running the `check` task
@@ -481,16 +511,32 @@ koverReport {
         xml {
             // overriding filters for default xml report
             filters {
-              excludes {  }
-              includes {  }
+                excludes {  }
+                includes {  }
+            }
+        }
+      
+        html {
+            // overriding filters for default HTML report
+            filters {
+                excludes {  }
+                includes {  }
+            }
+        }
+      
+        ic {
+            // overriding filters for default IC report
+            filters {
+                excludes {  }
+                includes {  }
             }
         }
       
         log {
             // overriding filters for logging report
             filters {
-              excludes {  }
-              includes {  }
+                excludes {  }
+                includes {  }
             }   
         }
     }  
@@ -508,17 +554,33 @@ koverReport {
         xml {
             // overriding filters for xml report for `release` build variant
             filters {
-              excludes {  }
-              includes {  }
+                excludes {  }
+                includes {  }
+            }
+        }
+      
+        html {
+            // overriding filters for HTML report for `release` build variant
+            filters {
+                excludes {  }
+                includes {  }
+            }
+        }
+      
+        ic {
+            // overriding filters for IC report for `release` build variant
+            filters {
+                excludes {  }
+                includes {  }
             }
         }
 
         log {
-          // overriding filters for logging report for `release` build variant
-          filters {
-            excludes {  }
-            includes {  }
-          }
+            // overriding filters for logging report for `release` build variant
+            filters {
+                excludes {  }
+                includes {  }
+            }
         }
     }
 }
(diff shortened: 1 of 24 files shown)


docs/gradle-plugin/configuring.md
```diff
@@ -93,6 +93,36 @@ koverReport {
             }
         }
 
+        // configure IC report
```
sandwwraith, line 96: Here and below:
it is extremely confusing to see 'IC' letters without knowing what they mean. So in comments you have to use 'Intellij Coverage binary report' everywhere. However, it seems kinda of internal feature for me. Who needs this binary report anyway, except some JB tools? Does it have publicly accessible definition/specification somewhere? If it does not, I suggest implementing the feature without advertising it — regular KDoc should be enough for internal-oriented feature.

docs/gradle-plugin/index.md
```diff
@@ -91,7 +91,7 @@ Using JVM projects and Android projects differs, use the recommendations below a
 ### Kotlin JVM project
 These are projects that use Kotlin Gradle plugin.
 
-For such projects, only the [Kover default tasks](#kover-default-tasks) for generating reports are used: `koverHtmlReport`, `koverXmlReport`, `koverVerify`.
+For such projects, only the [Kover default tasks](#kover-default-tasks) for generating reports are used: `koverHtmlReport`, `koverXmlReport`, `koverIcReport`, `koverVerify`.
```
sandwwraith, line 94: I suggest to use `koverBinaryReport` name for such tasks, as it is at least more descriptive. Probably configuration block should be changed to `binary` as well.

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/reports/ReportsVariantApplier.kt
```diff
@@ -135,6 +144,14 @@ internal sealed class ReportsVariantApplier(
             runOnCheck += xmlTask
         }
 
+        icTask.configure {
+            icFile.convention(reportConfig.ic.icFile)
+            filters.set((reportConfig.ic.filters ?: reportConfig.filters ?: commonFilters).convert())
+        }
+        runOnCheck += reportConfig.ic.onCheck.map { run ->
```
sandwwraith, line 151: Why we have to do `map` here while `reportConfig.xml.onCheck` can be accessed directly?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tools/kover/IcReport.kt
```diff
@@ -0,0 +1,51 @@
...
+}
+
+internal interface IcReportParameters : ReportParameters {
+    val icFile: Property<File>
+}
+
+internal abstract class IcReportAction : WorkAction<IcReportParameters> {
+    override fun execute() {
+        val icFile = parameters.icFile.get()
+        val smapFile = icFile.parentFile.resolve(icFile.nameWithoutExtension + ".smap")
```
sandwwraith, line 36: So we always generate `.smap` file next to `.ic` file? Shouldn't it be documented somewhere?

shanshin: So far, yes, but in the future, we may disable the creation of this file

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/reports/ReportsVariantApplier.kt
shanshin: Because a lazy configuration is used here, which is based on the use of providers.

Gradle [recommends](https://docs.gradle.org/current/userguide/lazy_configuration.html#creating_property_provider) using the `Provider#map` function to create them.

> You should prefer using map() instead, as this has some useful benefits, which we will see later.

In fact, this has many advantages.

docs/gradle-plugin/configuring.md
shanshin: The reports generated here can be used in `kover-cli` to generate human-readable reports.

In general, I agree, it may be worth renaming the block to `binary {`

docs/gradle-plugin/index.md
shanshin: I think `koverBinaryReport` is OK

docs/gradle-plugin/configuring.md
shanshin: I also agree that this description can be removed from here, because this is a very rare type of report that is needed in rare cases (which can be described in more detail later with the development of `kover-cli`)

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/reports/ReportsVariantApplier.kt
sandwwraith: Why `reportConfig.xml.onCheck` is not lazy then?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverNames.kt
```diff
@@ -60,6 +66,13 @@ public object KoverNames {
         return htmlReportTaskName(buildVariant)
     }
 
+    /**
+     * Name of the binary report generation task for [buildVariant] Android build variant for Android projects.
+     */
+    public fun androidIcReport(buildVariant: String): String {
```
sandwwraith, line 72: This one still has `ic`

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverNames.kt
```diff
@@ -126,6 +147,15 @@ public fun TaskContainer.koverAndroidHtmlReportName(buildVariantName: String): S
     return KoverNames.androidHtmlReport(buildVariantName)
 }
 
+/**
+ * Name of the binary report generation task for [buildVariantName] Android build variant for Android projects.
+ *
+ * Returns the same value as [KoverNames.androidIcReport].
+ */
+public fun TaskContainer.koverAndroidIcReportName(buildVariantName: String): String {
```
sandwwraith, line 155: And this too

sandwwraith (approved):

PR #427, opened 2023-07-10
Updates KDoc for `KoverReportExtension.xml` to be documented as `xml` configuring, not `html`

jordanterry:
Whilst reading the Kdoc I noticed that the docs for the `KoverReportExtension.xml` was incorrectly documented as configuring `html` reports.

I've opened the PR, but haven't been through every step of the contribution guidelines. I will do it shortly.


jordanterry:
I've not created an issue for this. Should I?

shanshin:
> I've not created an issue for this. Should I?

Hi, no, for such a small correction of the documentation, this is not necessary

jordanterry:
@shanshin not a problem, I've found the documentation around Kover excellent :D

PR #432, opened 2023-07-12
Fixed usage Kover Gradle Plugin in buildSrc directory

### LINKED ISSUE #415: Compile error in version `0.7.1`
**Describe the bug**
I'm using the Kover Gradle plugin in my `buildSrc` folder : 

```
dependencies {
    implementation("org.jetbrains.kotlinx", "kover-gradle-plugin", "0.7.1")
}
```

It works fine with version `0.7.0`, but I get a compile error with version `0.7.1`.

**Errors**
```
 Class 'kotlinx.kover.api.KoverMergedConfig' was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.8.0, expected version is 1.5.1.
```

**Expected behavior**
The Gradle task `:buildSrc:compileKotlin` should not fail.

**Reproducer**
Including the dependency to Kover Gradle plugin inside the `buildSrc` folder fails the build.

**Reports**
N/A

**Environment**
 - Kover Gradle plugin version: 0.7.1
 - Kotlin Gradle plugin: 1.8.10
 - Gradle version: 7.3.1
 - Kotlin project type: Kotlin/Android
 - Coverage Toolset (if customized in build script): Kover (no extra config)

shanshin:
Fixes #415

`kotlin-dsl` itself specifies the language version to ensure compatibility of the Kotlin DSL API. 
Since we ourselves guarantee and test compatibility with previous Gradle versions, we can override language version.
The easiest way to do this now is to specify the version in the `afterEvaluate` block

### DIFF
--- a/kover-gradle-plugin/build.gradle.kts
+++ b/kover-gradle-plugin/build.gradle.kts
@@ -1,3 +1,5 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 import java.time.LocalDate
 import java.time.format.DateTimeFormatter
 import org.jetbrains.kotlin.gradle.plugin.KotlinCompilation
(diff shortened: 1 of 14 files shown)


sandwwraith (approved):

sandwwraith:
Where do you set `kover.test.gradle.version`? On TeamCity?

shanshin:
> Where do you set `kover.test.gradle.version`? On TeamCity?

This parameter is optional, if it is not specified, Gradle is used, which builds the project itself.

In teamcity, in the future, it is necessary to configure builds with different versions of Gradle - this is planned for about version `0.9.0`

PR #444, opened 2023-08-03
Replaced absolute paths in the Kover artifact with relative ones

### LINKED ISSUE #440: koverGenerateArtifact task generates a cacheable report that contains absolute path
**Describe the bug**
`koverGenerateArtifact` task generates a report that contains absolute paths. That task is cacheable which means that subsequent builds pulling from the cache will get the same task output. The problem is that absolute paths are not the same on different machines, leading to subsequent tasks (e.g. koverXmlReport) to process invalid paths on subsequent builds/machines and as a result generates an empty report

**Expected behavior**
`koverGenerateArtifact` should generate the content of its output as relative paths to the root of the project. This should be an easy fix. 

**Reports**
Example of the output file from koverGenerateArtifact stored in our build cache:
```
/var/lib/buildkite-agent/builds/xxx/android-diff/build-src/lib-android-lint/src/main/kotlin
/var/lib/buildkite-agent/builds/xxx/android-diff/build-src/lib-android-lint/src/main/java

/var/lib/buildkite-agent/builds/xxx/android-diff/build-src/lib-android-lint/build/classes/java/main
/var/lib/buildkite-agent/builds/xxx/android-diff/build-src/lib-android-lint/build/classes/kotlin/main

/var/lib/buildkite-agent/builds/xxx/android-diff/build-src/lib-android-lint/build/kover/bin-reports/test.ic
```

shanshin:
Fixes #440

### DIFF
--- a/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/BuildCacheRelocationTests.kt
+++ b/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/BuildCacheRelocationTests.kt
@@ -12,7 +12,7 @@ import kotlin.test.assertEquals
 
 class BuildCacheRelocationTests {
     @Test
-    fun testDefaultTasks() {
+    fun test() {
         val cachePath = Files.createTempDirectory("test-gradle-cache-").toFile().canonicalPath
 
         val cachePatch = """
@@ -33,6 +33,9 @@ class BuildCacheRelocationTests {
         assertEquals("SUCCESS", result1.taskOutcome(":koverHtmlReport"))
         assertEquals("SUCCESS", result1.taskOutcome(":koverBinaryReport"))
         assertEquals("SUCCESS", result1.taskOutcome(":koverVerify"))
+
+        // delete the previous build to simulate project relocation
+        gradleBuild1.targetDir.deleteRecursively()
 
         /*
         since the build is created from a template,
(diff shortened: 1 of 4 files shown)


kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/BuildCacheRelocationTests.kt
```diff
@@ -42,12 +50,17 @@ class BuildCacheRelocationTests {
...
+        val result2 = gradleBuild2.runWithParams("koverXmlReport", "koverHtmlReport", "koverBinaryReport", "koverVerify", "--build-cache", "--info")
+        try {
+            assertEquals("FROM-CACHE", result2.taskOutcome(":test"))
+            assertEquals("FROM-CACHE", result2.taskOutcome(":koverGenerateArtifact"))
+            assertEquals("FROM-CACHE", result2.taskOutcome(":koverXmlReport"))
+            assertEquals("FROM-CACHE", result2.taskOutcome(":koverHtmlReport"))
+            assertEquals("FROM-CACHE", result2.taskOutcome(":koverBinaryReport"))
+            assertEquals("FROM-CACHE", result2.taskOutcome(":koverVerify"))
+        } catch (e: Exception) {
+            throw AssertionError("Build log \n${result1.output}",e)
```
sandwwraith, line 62:
```suggestion
            throw AssertionError("Build log \n${result2.output}",e)
```

sandwwraith (approved):

Issue #437, opened 2023-07-26
Kover doesn't work with configuration cache in Gradle 8.3-rc-1


NikolayMetchev:
**Describe the bug**
When running with gradle 8.3-rc-1 configuration cache breaks.

**Errors**
```
Execution optimizations have been disabled for task ':services:pes:pes-admin-service:test' to ensure correctness due to the following reasons:
  - Type 'org.gradle.api.tasks.testing.Test' property 'jvmArgumentProviders' with nested type 'kotlinx.kover.gradle.plugin.appliers.JvmTestTaskArgumentProvider' is not supported. Reason: Nested types are expected to either declare some annotated properties or some behaviour that requires capturing the type as input. For more information, please refer to https://docs.gradle.org/8.3-rc-1/userguide/validation_problems.html#unsupported_nested_type in the Gradle documentation.
  - Type 'org.gradle.api.tasks.testing.Test' property 'toolVariant' with nested type 'kotlinx.kover.gradle.plugin.tools.KoverToolDefaultVariant' is not supported. Reason: Nested types are expected to either declare some annotated properties or some behaviour that requires capturing the type as input. For more information, please refer to https://docs.gradle.org/8.3-rc-1/userguide/validation_problems.html#unsupported_nested_type in the Gradle documentation.
```

**Expected behavior**
Configuration cache to continue working

**Reproducer**

**Reports**
If applicable, report files or screenshots.

**Environment**
 - Kover Gradle Plugin version: 0.7.2
 - Gradle version: 8.3-rc-1
 - Kotlin project type: Kotlin/JVM
 - Coverage Toolset default
 - Other context important for this bug: Mac OS

shanshin:
Relates https://github.com/gradle/gradle/issues/26018.
Should be fixed on the Gradle side.

marukami:
Seems a fix was merged https://github.com/gradle/gradle/pull/26034.
When I tried 8.3-20230808000007+0000 it seem to fix the issue for me at least.

shanshin:
I confirm, the warning has disappeared in `8.3-rc-4`.

PR #449, opened 2023-08-15
Added functional test on Gradle nested classes validation error

shanshin:
Relates #437

### DIFF
--- a/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/GradleValidationTests.kt
+++ b/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/GradleValidationTests.kt
@@ -1,0 +1,16 @@
+package kotlinx.kover.gradle.plugin.test.functional.cases
+
+import kotlinx.kover.gradle.plugin.test.functional.framework.checker.CheckerContext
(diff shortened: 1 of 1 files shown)


kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/GradleValidationTests.kt
```diff
@@ -0,0 +1,16 @@
+package kotlinx.kover.gradle.plugin.test.functional.cases
+
+import kotlinx.kover.gradle.plugin.test.functional.framework.checker.CheckerContext
+import kotlinx.kover.gradle.plugin.test.functional.framework.starter.TemplateTest
+import kotlin.test.assertFalse
+
+internal class GradleValidationTests {
+
+    // test on validation error https://github.com/gradle/gradle/issues/26018
```
sandwwraith, line 9: I assume this is a bug fixed in Gradle, why this test is needed?

shanshin: In case this error is reproduced in the future in Gradle, or incorrect `@Nested` classes are added in the plugin

sandwwraith (approved):

PR #469, opened 2023-09-22
Disabled writing of FreeMarker logs to stdout when generating an HTML report

### LINKED ISSUE #446: Intermittent error output: "Error executing FreeMarker template"
**Describe the bug**

Randomly every so often I see the output `Error executing FreeMarker template`. I'm not 100% sure it is from kover, but maybe 98% sure. There is no other output, and it is printed to standard error. 

Upon searching for this output, I see that FreeMarker is in fact used in IntelliJ's (and thus Kover's) code coverage. Also, I never saw this output before I started using Kover myself.

Code coverage seems to otherwise work completely as expected. 

**Errors**

`Error executing FreeMarker template` in red (standard error). That's it.

**Expected behavior**

As I said above code coverage seems to still proceed normally. 

If the FreeMarker error is handled internally, than the error should not be shown to the user as it is non-actionable, unless the user's log level is high. My `org.gradle.logging.level` is set to `warn`, so I don't expect to see any output unless it has a purpose of informing me of an issue.

If the error is not completely handled, and is in fact something that we should be concerned about, then there should be a full stack trace and a more informative message.

**Reproducer**

I cannot reproduce it. It happens at random. If there was a full stack trace or a more informative message I may be able to figure out how to reproduce it.

**Environment**
 - Kover Gradle Plugin version: 0.7.3
 - Gradle version: 8.2.1
 - Kotlin project type: Kotlin/Multiplatform + Android
 - Coverage Toolset: Kover
 - OS version: macOS 13.4.1
 
 See the very place that this message is logged in FreeMarker [here](https://github.com/apache/freemarker/blob/df938ce6120ca155b87d48df97b3a6d62123b17f/src/main/java/freemarker/core/Environment.java#L1207)

shanshin:
Due to an untraceable error of FreeMarker templates, repeated attempts to generate a report are triggered in some cases. In case of unsuccessful attempts, FreeMarker writes to its logger, which is sent to stdout by default. 
If the error repeats frequently, an exception will be thrown with an error message - so we can ignore FreeMarker's own logging when generating an HTML report.

Fixes #446


fzhinkin (approved):
LGTM

sandwwraith (approved):

PR #471, opened 2023-09-24
Update HTML report path to be clickable

### LINKED ISSUE #470: HTML report paths are not clickable
**Describe the bug**
HTML report paths can't be clicked on in the output of IntelliJ / Android Studio / etc, and instead need to be manually copy and pasted into the browser

Ezard:
Fixes HTML report paths not being clickable

Closes #470

### DIFF
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tasks/reports/KoverHtmlTask.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tasks/reports/KoverHtmlTask.kt
@@ -4,6 +4,7 @@
 
 package kotlinx.kover.gradle.plugin.tasks.reports
 
+import java.io.File
 import org.gradle.api.file.*
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.*
@@ -28,7 +29,7 @@ internal abstract class KoverHtmlTask : AbstractKoverReportTask() {
     }
 
     fun printPath(): Boolean {
-        logger.lifecycle("Kover: HTML report for '$projectPath' file://${reportDir.get().asFile.canonicalPath}/index.html")
+        logger.lifecycle("Kover: HTML report for '$projectPath' file://${reportDir.get().asFile.canonicalPath}${File.separator}index.html")
         return true
     }
 }
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tasks/reports/KoverHtmlTask.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tasks/reports/KoverHtmlTask.kt
@@ -4,10 +4,11 @@
 
 package kotlinx.kover.gradle.plugin.tasks.reports
 
-import java.io.File
 import org.gradle.api.file.*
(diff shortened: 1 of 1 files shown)


kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tasks/reports/KoverHtmlTask.kt
```diff
@@ -28,7 +30,14 @@ internal abstract class KoverHtmlTask : AbstractKoverReportTask() {
     }
 
     fun printPath(): Boolean {
-        logger.lifecycle("Kover: HTML report for '$projectPath' file://${reportDir.get().asFile.canonicalPath}/index.html")
+        val clickablePath = URI(
```
shanshin, line 33: It looks like such code gives the same result:

```
        val clickablePath = URI(
            "file",
            "",
            file.resolve("index.html").canonicalPath,
            null
        ).toASCIIString()
```

shanshin: Perhaps it's even better to try like this:
```
        val clickablePath = URI(
            "file",
            "",
            reportDir.get().file("index.html").asFile.toURI().toASCIIString(),
            null
        ).toASCIIString()
```

Ezard: @shanshin I agree with the first one - the second one crashes for me, because the path ends up being something like `C;/path/to/file.html`, but `URI` requires a leading slash - `canonicalPath` is the bit that adds this slash in

Ezard: Actually, the first one appears to crash for me too 🤔

Ezard:
```kotlin
reportDir.get().asFile.resolve("index.html").toURI().path
```
seems to work, as an alternative to
```kotlin
File(reportDir.get().asFile.canonicalPath, "index.html").toURI().path
```

shanshin: `reportDir.get().asFile.resolve("index.html").toURI().path`
Ok, this is also a good variant if it works stably on Windows

Ezard: Found some issues with that version on Windows :(

My recommendation would be to just use the original version in this PR - I took that straight from the Gradle codebase, so I assume that that's very battle-tested and reliable

shanshin: Ok. 
Could you please specify what problems have appeared? 
This may come in handy in the future in other places

shanshin (approved):

shanshin:
Thanks for the effort!

PR #490, opened 2023-10-13
Fixed adding data to an already existing binary report file

### LINKED ISSUE #489: Jacoco coverage report won't reflect latest change
**Describe the bug**
Jacoco report was not update correctly after delete one of the tests.
Coverage report still showing object is covered.

**Errors**
-

**Expected behavior**
Coverage report should show object is not covered.

**Reproducer**
Basic project with two test file.

build.gradle.kts
```
plugin {
  id("org.jetbrains.kotlinx.kover") version "0.7.4"
}
...
kover {
    useJacoco("0.8.10")
}
```
Target class
```
object HelloWorld {
    fun hello() = "Hello"
    fun world() = "${hello()} World"
}
```
Test file 1
```
class HelloTest : DescribeSpec({
    describe("hello") {
        context("hello") {
            it("should return hello") {
                HelloWorld.hello() shouldBe "Hello"
            }
        }
    }
})
```
Test file 2
```
import io.kotest.core.spec.style.DescribeSpec
import io.kotest.matchers.shouldBe

class WorldTest : DescribeSpec({

    describe("world") {
        context("world") {
            it("should return hello world") {
                HelloWorld.world() shouldBe "Hello World"
            }
        }
    }

})
```

* trigger koverHtmlReport task, I could see HelloWorld class is covered 100%.
* delete test file 2, trigger koverHtmlReport task, HelloWorld class still covered 100%.

**Reports**
-

**Environment**
 - Kover Gradle Plugin version: 0.7.4
 - Gradle version: 7.6.1
 - Kotlin project type: Kotlin
 - Coverage Toolset (if customized in build script): JaCoCo("0.8.10")
 - Other context important for this bug: mac

shanshin:
If the build cache is not used and the build directory is not cleared before the test run, then the coverage data from previous test runs are merged with the data from the recent run. In this case, the report may include coverage from those tests that were removed in the latest version of the code.

To solve this, it is necessary to delete the binary report file before each run of the test task.

Fixes #489

### DIFF
--- a/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/ReportsCachingTests.kt
+++ b/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/ReportsCachingTests.kt
@@ -3,6 +3,7 @@
  */
 package kotlinx.kover.gradle.plugin.test.functional.cases
 
+import kotlinx.kover.gradle.plugin.commons.CoverageToolVendor
 import kotlinx.kover.gradle.plugin.test.functional.framework.configurator.*
 import kotlinx.kover.gradle.plugin.test.functional.framework.starter.*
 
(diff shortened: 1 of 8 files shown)


kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/ReportsCachingTests.kt
```diff
@@ -29,55 +30,80 @@ internal class ReportsCachingTests {
...
-        reportAndCheck("SUCCESS", true)
+        reportAndCheck("SUCCESS", cached = true)
 
         edit("src/test/kotlin/TestClass.kt") {
             "$it\n class AdditionalTest"
         }
 
-        // tasks must be restarted after tests are edited
-        reportAndCheck("SUCCESS", true)
+        // test task must be restarted after test class is edited, but reports is up-to-date because no sources changed
```
sandwwraith, line 61: What happens when I edit the test class in a way that I delete some tests so the resulting coverage (of unchanged sources) can be lower? Will it trigger validation checks, then? Are you sure this is the correct behavior?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/JvmTestTaskConfigurator.kt
```diff
@@ -39,6 +39,12 @@ internal class JvmTestTaskConfigurator(
             }
         testTask.dependsOn(data.findAgentJarTask)
 
+        testTask.doFirst {
+            // delete report so that when the data is re-measured, it is not appended to an already existing file
+            // see https://github.com/Kotlin/kotlinx-kover/issues/489
+            binReportProvider.get().asFile.delete()
```
sandwwraith, line 45: Does it matter if it was in the build cache or not?

kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/ReportsCachingTests.kt
shanshin: Changing the coverage will cause the report to be regenerated.
In this case, a test class is added in which there is not a single test - therefore, this does not change the coverage

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/JvmTestTaskConfigurator.kt
shanshin: No difference, because `doFirst` is executed only when the Gradle task is running (it has the SUCCESS status).
If the task has the status of `UP-TO-DATE` or `FROM-CACHE`, then `doFirst` is not called

sandwwraith (approved):

PR #502, opened 2023-11-06
Fixed applying of the root verification rules

### LINKED ISSUE #459: Top level verification rule is not applied to verification tasks for Android build variants
**Describe the bug**

From descriptions in the docs, top level verification rule should be applied to all variants including Android build variants:

- https://kotlin.github.io/kotlinx-kover/gradle-plugin/configuring#configuring-android-reports

> // verification rules for verification tasks in all variants

- https://kotlin.github.io/kotlinx-kover/gradle-plugin/configuring#verification-rules

> Rules can be specified for all report variants:
>
> ```
> koverReport {
>     verify {
>         rule {
>             minBound(50)
>         }
>     }
> }
> ```

But it seems that only default verification task (i.e. `koverVerify`) verifies coverage report based on the top level rule and Android report verification tasks (i.e. `koverVerify{VariantName}`) do not use the rule.

**Errors**

No error is found.

**Expected behavior**

Verification rule for all report variants is applied to Android report variants, which are executed by `koverVerify{VariantName}`, as well as default report.

**Reproducer**

I created a reproducer project: https://github.com/nashcft/repro-kover-verification

**Reports**

Following screenshot shows the result of verification task executions on the reproducer project mentioned above.  Note that the verification for `release` variant in the capture is failed because the variant has variant-specific verification rule.

Screenshot 2023-09-17 at 13 22 36

**Environment**
 - Kover Gradle Plugin version: 0.7.3
 - Gradle version: 8.2
 - Kotlin project type: confirmed with Kotlin/Android and  Kotlin/Multiplatform + Android
 - Coverage Toolset (if customized in build script): Kover
 - Other context important for this bug:
   - OS: macOS Ventura 13.5.2

shanshin:
Now, when creating the Android report configuration, on Check is immediately set for verification.

This is not necessary, because if not specified, only the `koverVerify` task is started when executing check, but not the tasks of verifying Android build variant. This has a negative side effect, because the verification for the build variant is always specify, and it will override the common rules from the root of the report config `koverReport { verify { /* common rules */ }  }`

Fixes #459

### DIFF
--- a/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/VerificationTests.kt
+++ b/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/VerificationTests.kt
@@ -5,8 +5,15 @@
 package kotlinx.kover.gradle.plugin.test.functional.cases
 
 import kotlinx.kover.gradle.plugin.dsl.*
+import kotlinx.kover.gradle.plugin.test.functional.framework.checker.CheckerContext
+import kotlinx.kover.gradle.plugin.test.functional.framework.checker.createCheckerContext
 import kotlinx.kover.gradle.plugin.test.functional.framework.configurator.*
+import kotlinx.kover.gradle.plugin.test.functional.framework.runner.buildFromTemplate
+import kotlinx.kover.gradle.plugin.test.functional.framework.runner.runWithParams
 import kotlinx.kover.gradle.plugin.test.functional.framework.starter.*
+import org.junit.jupiter.api.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
 
 internal class VerificationTests {
     @SlicedGeneratedTest(allLanguages = true, allTools = true)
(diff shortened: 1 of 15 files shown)


sandwwraith:
> onCheck is immediately set for verification.

Is immediately set to `false`, judging by the edited code?

> This is not necessary, because if not specified

If what is not specified?

> because the verification for the build variant is always specify

Is specified by default, you mean?

I'm not sure I've understood the fix completely: #459 was about the rules from default `koverReport { verify { ... }}` being ignored when running `koverVerify{VariantName}`. How exactly setting `onCheck` to true will fix this issue? Now `koverVerify{VariantName}` will also run standard `koverVerify`? Or `check` task will run all `koverVerify{VariantName}`?

shanshin:
> How exactly setting onCheck to true will fix this issue?

No, `onCheck`  does not become true, it stops being filled in at all. 
According to the rules for Android reports, if `onCheck` is not specified, it is [taken as false](https://github.com/Kotlin/kotlinx-kover/blob/59c6718a40db5c5344f0e08c8d576f44151caacd/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/reports/ReportsVariantApplier.kt#L174C31-L174C31).
 

Now, to install onCheck, the `reports.verify {` block is called.
if you look at its implementation, you can see that the default value for `verify` config is [null](https://github.com/Kotlin/kotlinx-kover/blob/59c6718a40db5c5344f0e08c8d576f44151caacd/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/internal/KoverReportExtension.kt#L25). 
When a user writes `verify { }`, according to the current rules, it automatically [resets all filters](https://github.com/Kotlin/kotlinx-kover/blob/59c6718a40db5c5344f0e08c8d576f44151caacd/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/internal/KoverReportExtension.kt#L38) declared in the [hierarchy above](https://kotlin.github.io/kotlinx-kover/gradle-plugin/configuring#reports-filtering).

The problem was that the `onCheck` setting took place through a `verify` block call, which should be filled in by the user, so the filters seem to be erased.

sandwwraith (approved):
This explanation makes much more sense, thank you. Can you add it to the commit message?

PR #516, opened 2023-11-30
Update Gradle Dokka configuration to make sure "source" button is visible in all API docs

whyoleg:
For context: https://github.com/Kotlin/kotlinx.coroutines/issues/3929 and https://github.com/Kotlin/kotlinx.coroutines/pull/3960

BTW, is there a reason why `kover` API reference is not published under `https://kotlinlang.org/docs/home.html` API reference section?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ out/
 # Gradle files
 build/
 .gradle/
+
+# generated dokka documentation
+/docs/gradle-plugin/dokka/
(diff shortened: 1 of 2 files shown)


whyoleg:
Simplified configuration based on: https://github.com/Kotlin/kotlinx-datetime/pull/323#discussion_r1412147614

Could you also help to understand how documentation is published and so where `remoteUrl` should be pointed, so it would be stable for current releases version?

shanshin (approved):

shanshin:
> Could you also help to understand how documentation is published and so where remoteUrl should be pointed, so it would be stable for current releases version?

Documentation is now being generated and published manually, https://kotlin.github.io/kotlinx-kover/gradle-plugin/dokka/

The documentation for the stable version is currently under development and it has not yet been decided how and where it will be published. In the coming months, we will continue to publish as is

PR #539, opened 2024-02-14
Added the ability to specify a header for an XML report

### LINKED ISSUE #527: Set XML report name to be the module name

Currently, the XML coverage report generated by Kover, when using the default engine, has its name set to `Intellij Coverage Report`, ie:
```xml
<report name="Intellij Coverage Report">
```

When using JaCoCo, the name is set to the module's name (for example `:app`).

Would it be possible to have the same behaviour for both engines? The report name may be used by other tools, for example https://github.com/marketplace/actions/jacoco-report. Having a unique/identifiable name for each report may then be useful.

It would be great if Kover's default engine could use the module's name for the report name, instead of `Intellij Coverage Report`.

shanshin:
Resolves #527

### DIFF
--- a/docs/gradle-plugin/configuring.md
+++ b/docs/gradle-plugin/configuring.md
@@ -94,6 +94,9 @@ koverReport {
             //  generate an XML report when running the `check` task
             onCheck = false
 
+            // XML report title (the location depends on the library)
+            title.set("Custom XML report title")
+          
             // XML report file
             setReportFile(layout.buildDirectory.file("my-project-report/result.xml"))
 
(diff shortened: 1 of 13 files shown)


sandwwraith:
Shouldn't this PR be done after merging DSLv4 to avoid conflicts?

shanshin:
> Shouldn't this PR be done after merging DSLv4 to avoid conflicts?

No, there won't be many conflicts.
DSL change is a serious update, first a Beta release will be released, which will be difficult for users to quickly switch to.

sandwwraith (approved):

PR #541, opened 2024-02-14
Fixed the presence of classes instrumented with Robolectric in Jacoco reports

### LINKED ISSUE #530: Kotlin test coverage not reported for Java production code with JaCoCo
This is my configuration:

```
plugins {
}

val configuration = extensions.create("glscoverage", GlsCoverageExtension::class.java)

afterEvaluate {
    kover {
        // not working
        useJacoco("0.8.11")
    }

    val expected = configuration.expectedCoverage.get()
    val threshold = configuration.coverageThreshold.get()
    val minBound = (expected - threshold).coerceAtLeast(0)
    val maxBound = (expected + threshold).coerceAtMost(100)

    koverReport {
        defaults {
            mergeWith(configuration.buildVariant.get())
        }

        filters {
            excludes {
                annotatedBy("*Generated*")
                classes(configuration.classFilter.get())
            }
        }

        verify {
            rule {
                minBound(minBound)
                maxBound(maxBound)
            }
        }

        androidReports(configuration.buildVariant.get()) {}
    }

    tasks.withType(Test::class.java).configureEach {
        finalizedBy(tasks.named("koverXmlReport"))
    }
}
```

Now I have separate Kotlin and Java source sets for my mixed-language module configured, i.e. `src/{main,test}/{kotlin,java}` and I have a Kotlin test that tests various Java source files. With the above configuration the coverage for these Java source files is reported as 0, even if I move the test from `src/test/kotlin` to `src/test/java`, however, if I comment out `useJacoco` above and let the IJ coverage processor do it's thing, the coverage of the classes in question is properly reported.

**Environment**
 - Kover Gradle Plugin version: 0.7.5
 - Gradle version: 8.5
 - Kotlin project type: Kotlin/Android
 - Coverage Toolset (if customized in build script): Kover, Jacoco (0.8.11)

shanshin:
It seems that after Robolectric instrumentation, source locations are lost and JaCoCo does not instrument such classes by default. There is a need to change the flag to disable this behavior.

After changing this flag, errors occur in the instrumentation of JVM classes (because they were previously excluded due to the lack of sources). The package with these classes was determined empirically.

Fixes #530

### DIFF
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tools/jacoco/JacocoOnlineInstrumentation.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tools/jacoco/JacocoOnlineInstrumentation.kt
@@ -8,7 +8,7 @@ import java.io.*
 
 internal fun buildJvmAgentArgs(jarFile: File, binReportFile: File, excludedClasses: Set<String>): List<String> {
     val agentArgs = listOfNotNull(
-        "destfile=${binReportFile.canonicalPath},append=true,inclnolocationclasses=false,dumponexit=true,output=file,jmx=false",
+        "destfile=${binReportFile.canonicalPath},append=true,inclnolocationclasses=true,dumponexit=true,output=file,jmx=false",
         excludedClasses.joinToFilterString("excludes")
     ).joinToString(",")
 
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/JvmTestTaskConfigurator.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/JvmTestTaskConfigurator.kt
@@ -45,8 +45,14 @@ internal class JvmTestTaskConfigurator(
             binReportProvider.get().asFile.delete()
         }
 
-        // Always excludes android classes, see https://github.com/Kotlin/kotlinx-kover/issues/89
-        val excluded = data.excludedClasses + listOf("android.*", "com.android.*")
+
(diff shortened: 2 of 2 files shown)


kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tools/jacoco/JacocoOnlineInstrumentation.kt
```diff
@@ -8,7 +8,7 @@ import java.io.*
 
 internal fun buildJvmAgentArgs(jarFile: File, binReportFile: File, excludedClasses: Set<String>): List<String> {
     val agentArgs = listOfNotNull(
-        "destfile=${binReportFile.canonicalPath},append=true,inclnolocationclasses=false,dumponexit=true,output=file,jmx=false",
+        "destfile=${binReportFile.canonicalPath},append=true,inclnolocationclasses=true,dumponexit=true,output=file,jmx=false",
```
sandwwraith, line 11: Shouldn't we enable this only if Robolectric was used? For normal projects, you likely do not need to include classes without locations

sandwwraith (approved):

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tools/jacoco/JacocoOnlineInstrumentation.kt
shanshin: All classes will not be included in the report anyway.
We don't know for sure whether Robolectric alone causes such an error or not.

It is quite non-trivial to understand before run whether such instrumentation occurred using Robolectric (because even analyze the dependencies of the Gradle project)

PR #563, opened 2024-03-13
Changed log level for print coverage task

### LINKED ISSUE #557: koverPrintCoverage does not work when org.gradle.logging.level=warn
**Describe the bug**

When `org.gradle.logging.level=warn`, `koverPrintCoverage` does not produce any output.

**Expected behavior**

`org.gradle.logging.level` is used to control logging. Logging is a passive way to keep information on problems, errors, and just general information of a process. When `org.gradle.logging.level=warn`, we are saying that we only want warnings; we only want actionable information kept.

`koverPrintCoverage` is not passive logging. It is an optional task that is triggered manually when we want to see coverage information. It should not require that we set `org.gradle.logging.level=lifecycle`. Unlike all of the other information that is printed at the `lifecycle` log level, `koverPrintCoverage` is information that we have manually triggered to see.

Compare `koverPrintCoverage` with `tasks`. If `org.gradle.logging.level=warn` and we run `./gradlew tasks`, we still see the printed output from the `tasks` task. `koverPrintCoverage` should work the same way.

shanshin:
At the moment, the coverage was printed with the `lifecycle` level. However, if user limit the logs to the `--warn` level, the coverage stops printing. 
Now coverage is printed with the quite level, thanks to this it is printed even with the limitations of `--warn` and `--quiet`.

Fixes #557

### DIFF
--- a/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/LoggingTaskTests.kt
+++ b/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/LoggingTaskTests.kt
@@ -10,6 +10,7 @@ import kotlinx.kover.gradle.plugin.dsl.GroupingEntityType
 import kotlinx.kover.gradle.plugin.dsl.CoverageUnit
 import kotlinx.kover.gradle.plugin.test.functional.framework.configurator.BuildConfigurator
 import kotlinx.kover.gradle.plugin.test.functional.framework.starter.GeneratedTest
+import kotlin.test.assertContains
 import kotlin.test.assertEquals
 
 internal class LoggingTaskTests {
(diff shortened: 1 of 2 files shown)


sandwwraith:
Documentation says that "-q, --quiet: Log errors only." I wouldn't expect to see coverage logs at this level unless low coverage is reason for failing the build. IMO, if user specifies `--quiet`, they expect log in the destination file only, not in the output.

shanshin:
> Documentation says that "-q, --quiet: Log errors only." I wouldn't expect to see coverage logs at this level unless low coverage is reason for failing the build. IMO, if user specifies --quiet, they expect log in the destination file only, not in the output.

According [Gradle docs](https://docs.gradle.org/current/userguide/logging.html) quiet level is a important message which should always be printed.

At the same time, if you do not need coverage logs, then you do not need to call the `koverLog` task.
Task `koverLog` is intended only for printing the coverage to the output, users don't even know about the existence of the file - it's an implementation detail.

PR #577, opened 2024-03-21
Implemented warn on verification error

### LINKED ISSUE #339: Warn on failed verification
Sometimes I may want the vertification task to simply produce a warning rather than a failure.

I suggest adding a configuration property to the buildscript dsl to allow users to control whether a failed verification produces a warning or an exception.

shanshin:
Added property to suppress task execution failure if there has been a violation of the verification rules

Resolves #339


kover-gradle-plugin/docs/configuring.md
```diff
@@ -29,11 +29,14 @@ koverReport {
   
     // verification rules for verification task
     verify {
+        // fail on verification error
+        warnOnFailure = false
```
sandwwraith, line 33: IMO, 'warnOnFailure' doesn't indicate that failure won't happen. I think something like `warningInsteadOfFailure` or conveys meaning better.

kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/VerificationTests.kt
```diff
@@ -175,6 +175,71 @@ Rule violated: lines missed count for package 'org.jetbrains.kover.test.function
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        run("koverVerify") {
+            taskOutput(":koverVerify") {
+                match {
```
sandwwraith, line 199: Should we also check that build finished successfully?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/tasks/VariantReportsSet.kt
```diff
@@ -117,6 +136,9 @@ internal class VariantReportsSet(
             onlyIf {
                 fileWithMessage.asFile.get().exists()
             }
+
+            // always execute
+            outputs.upToDateWhen { false }
```
sandwwraith, line 141: Why do we need to always execute this task? So it reports the same warning over and over?

kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/VerificationTests.kt
shanshin: Test will fail if build finished unsuccessfully

kover-gradle-plugin/docs/configuring.md
shanshin: Yeah, agreed

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/tasks/VariantReportsSet.kt
shanshin: So that the task behaves consistently.

If you call task twice in a row
```
% ./gradlew :koverVerify 
> Task :koverVerify 
Kover Verification Error
Rule violated: lines covered percentage is 10, but expected minimum is 79.0000
BUILD SUCCESSFUL in 8s

// since some changes

% ./gradlew :koverVerify 
BUILD SUCCESSFUL in 8s
```
There may be a mistaken impression that the error has gone away. However, the task was in `UP-TO-DATE` status because the source code files on which it depends were not changed.

sandwwraith (commented):
Did you push new commits? I see no changes since my last review

shanshin:
Sorry, pushed

sandwwraith (approved):

PR #578, opened 2024-03-21
Added DslMarker to Kover public interfaces

### LINKED ISSUE #572: 0.8.0-Beta DSL confusion between `minBound` and `minValue`.
I have spent a long time being very confused, stepping through debuggers, and creating a reproducer only to find out that I was using the DSL incorrectly. I think it is a design issue, as it is too easy for users to get mixed up.

I was doing this:

```kotlin
kover {
    reports {
        verify {
            rule {
                bound {
                    coverageUnits.set(CoverageUnit.BRANCH)
                    minBound(75)
                }
            }
        }
    }
}
```

And my output kept looking like this:

```txt
Rule violated: lines covered percentage is 12.500000, but expected minimum is 75
```

I was very confused why it kept using lines instead of branches. Finally, I discovered I needed to do this:

```kotlin
kover {
    reports {
        verify {
            rule {
                bound {
                    coverageUnits.set(CoverageUnit.BRANCH)
                    minValue = 75 // <-- Change here
                }
            }
        }
    }
}
```

And now it works:

```txt
Rule violated: branches covered percentage is 0.000000, but expected minimum is 75
```

I think that the correct solution here is to use the `@DslMarker` annotation on all DSL objects. I mean, create an annotation called `KoverDsl`, annotate that annotation with `@DslMarker`, and put `KoverDsl` on all kover dsl interfaces.

shanshin:
Fixes #572


sandwwraith (approved):
Does incorrect usage of DSL produce warning or error? If error, we probably can't make this change between 0.8.0-Beta and 0.8.0-Stable

shanshin:
> Does incorrect usage of DSL produce warning or error? If error, we probably can't make this change between 0.8.0-Beta and 0.8.0-Stable

The DSL may change slightly between `0.8.0-Beta` and `0.8.0` (because the goal of this beta version is to collect minimal feedback on the new DSL), so I don't see a problem with such changes in `Beta2`.

Also, such an error occurs only in kts files, there is no such check for Groovy.

Issue #570, opened 2024-03-18
Error trying to generate a single report using the newest beta version (0.8.0-Beta) 


igorromcy:
**Describe the bug**
I'm trying to use the latest Kover version since it has some great improvements around the single report approach and I'm facing this weird error when I'm trying to add a variant. Is that something inside the library or I might be missing some kind of configuration? 

**Errors**
```
Caused by: kotlinx.kover.gradle.plugin.commons.KoverCriticalException: Kover error: An error occurred while executing before Kover finalize action
Please create an issue in Kover Gradle Plugin repository: https://github.com/Kotlin/kotlinx-kover/issues
	at kotlinx.kover.gradle.plugin.appliers.FinalizeKoverKt.finalizing(FinalizeKover.kt:35)
	at kotlinx.kover.gradle.plugin.KoverGradlePlugin$apply$2.invoke(KoverGradlePlugin.kt:35)
	at kotlinx.kover.gradle.plugin.KoverGradlePlugin$apply$2.invoke(KoverGradlePlugin.kt:33)
	at
....

Caused by: org.codehaus.groovy.runtime.metaclass.MissingMethodExecutionFailed: No signature of method: kotlinx.kover.gradle.plugin.appliers.KoverMergeKt$wrap$3.addWithDependencies() is applicable for argument types: (ArrayList, Boolean) values: [[debug, fastdevDebug], false]
Possible solutions: addWithDependencies([Ljava.lang.String;, boolean)
Caused by: groovy.lang.MissingMethodException: No signature of method: kotlinx.kover.gradle.plugin.appliers.KoverMergeKt$wrap$3.addWithDependencies() is applicable for argument types: (ArrayList, Boolean) values: [[debug, fastdevDebug], false]
Possible solutions: addWithDependencies([Ljava.lang.String;, boolean)
```
**Code**
I'm applying the plugin to my build.gradle file 

``` 
    id("org.jetbrains.kotlinx.kover") version "0.8.0-Beta" apply true
``` 

and setting the config: 
``` 
kover {
    merge {
        allProjects()
        createVariant("coverage") {
            it.addWithDependencies(["debug", "fastDebug"], false)
        }
    }
}
``` 

**Environment**
 - Kover Gradle Plugin version: 0.8.0
 - Gradle version: 8.2.0
 - Kotlin project type: Kotlin/Multiplatform + Android

shanshin:
Hi,
try this way
```
kover {
    merge {
        allProjects()
[...]

PR #579, opened 2024-03-21
Added overload for functions of KoverVariantCreateConfig

shanshin:
Previous versions of the functions had vararg parameters, this is not always convenient if the List is presented when or Groovy is used.

Relates #570

### DIFF
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverVariantConfig.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverVariantConfig.kt
@@ -216,4 +216,21 @@ public interface KoverVariantCreateConfig: KoverVariantConfig {
      * If [optional] is `true` and a variant with given name is not found in the current project - in this case, the variant will not be searched even in dependencies.
      */
     public fun addWithDependencies(vararg variantNames: String, optional: Boolean = false)
+    /**
+     * Add to created variant classes, tests and instrumented classes from report variant with name [variantNames].
+     * This variant is taken only from the current project.
+     *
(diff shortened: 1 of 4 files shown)


kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverVariantConfig.kt
```diff
@@ -216,4 +216,21 @@ public interface KoverVariantCreateConfig: KoverVariantConfig {
      * If [optional] is `true` and a variant with given name is not found in the current project - in this case, the variant will not be searched even in dependencies.
      */
     public fun addWithDependencies(vararg variantNames: String, optional: Boolean = false)
+    /**
+     * Add to created variant classes, tests and instrumented classes from report variant with name [variantNames].
+     * This variant is taken only from the current project.
```
sandwwraith, line 221:
```suggestion
     * These variants are taken only from the current project.
```
Given that `Iterable` is usually some kind of collection, using plural form seems more logical

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverVariantConfig.kt
```diff
@@ -216,4 +216,21 @@ public interface KoverVariantCreateConfig: KoverVariantConfig {
...
+     *
+     * If [optional] is `false` and a variant with given name is not found in the current project, an error [KoverIllegalConfigException] is thrown.
+     */
+    public fun add(variantNames: Iterable<String>, optional: Boolean = false)
+
+    /**
+     * Add to created variant classes, tests and instrumented classes from report variant with name [variantNames].
+     * This variant is taken from the current project and all `kover(project("name"))` dependency projects.
```
sandwwraith, line 229: Same here

sandwwraith (approved):

PR #591, opened 2024-03-29
Expose reportDir property in KoverHtmlReport interface

### LINKED ISSUE #590: Expose `reportDir` property in `KoverHtmlReport` interface
if there is a need to analyze the HTML report after it is generated, or if you need to improve the report (for example, by editing CSS styles), you need to know the directory of each report.
Currently, it is not possible to easily link a specific report generation task and a directory.

Add property `reportDir` interface `kotlinx.kover.gradle.plugin.dsl.tasks.KoverHtmlReport`
```
interface KoverHtmlReport: KoverReport {
    val reportDir: Provider<Directory>
}
```

Also, use case described in [comment](https://github.com/Kotlin/kotlinx-kover/issues/588#issuecomment-2026408351).

shanshin:
if there is a need to analyze the HTML report after it is generated, or if you need to improve the report (for example, by editing CSS styles), you need to know the directory of each report. Currently, it is not possible to easily link a specific report generation task and a directory.

Resolves #590

### DIFF
--- a/kover-gradle-plugin/api/kover-gradle-plugin.api
+++ b/kover-gradle-plugin/api/kover-gradle-plugin.api
@@ -304,6 +304,7 @@ public abstract interface class kotlinx/kover/gradle/plugin/dsl/tasks/KoverBinar
 }
 
 public abstract interface class kotlinx/kover/gradle/plugin/dsl/tasks/KoverHtmlReport : kotlinx/kover/gradle/plugin/dsl/tasks/KoverReport {
+	public abstract fun getReportDir ()Lorg/gradle/api/provider/Provider;
 }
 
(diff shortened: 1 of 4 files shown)


sandwwraith (approved):
Why `KoverHtmlTaskConfig.htmlDir` is not suitable here?

shanshin:
> Why KoverHtmlTaskConfig.htmlDir is not suitable here?

because in this case, to add the `doFirst` action to the task, you first need to add the `variantName` property to `KoverReport` interfac (to match variant), and then redo `kover.reports.variant(task.variantName)` so that you can work with it as an object and call it during task execution - this method requires more effort and gives exactly the same result.

PR #595, opened 2024-04-09
Added property variantName to KoverReport interface

### LINKED ISSUE #587: Add property `variantName` to `KoverReport` interface
Property `variantName` will help to filter Kover tasks by selecting those that relate to specific variant.

So, to find the HTML report for `debug` Android build variant:
```
tasks.withType<KoverHtmlReport>().matching { 
    it.variantName == "debug" 
}
```

shanshin:
Property `variantName` will help to filter Kover tasks by selecting those that relate to specific variant.

Resolves #587

### DIFF
--- a/kover-gradle-plugin/src/functionalTest/templates/builds/android-test-tasks-filtering/app/build.gradle.kts
+++ b/kover-gradle-plugin/src/functionalTest/templates/builds/android-test-tasks-filtering/app/build.gradle.kts
@@ -1,0 +1,111 @@
+import kotlinx.kover.gradle.plugin.dsl.tasks.*
+
+plugins {
+    id ("org.jetbrains.kotlinx.kover")
+    id ("com.android.application")
+    id ("org.jetbrains.kotlin.android")
+}
+
+android {
+    namespace = "kotlinx.kover.test.android"
+    compileSdk = 32
+
+    defaultConfig {
+        applicationId = "kotlinx.kover.test.android"
+        minSdk = 21
+        targetSdk = 31
+        versionCode = 1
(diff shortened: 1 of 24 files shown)


kover-gradle-plugin/src/functionalTest/templates/builds/android-test-tasks-filtering/app/build.gradle.kts
```diff
@@ -0,0 +1,111 @@
...
+        println("HTML=$htmlName")
+        println("Verify=$verifyName")
+        println("Log=$logName")
+        println("Binary=$binaryName")
+    }
+}
+
+tasks.register("findTotalTasks") {
+    doLast {
+        val xmlName = tasks.withType<KoverXmlReport>().matching {
```
fzhinkin, line 85: Is it expected to be used frequently? Does it make sense to provide something like
```
fun <T : KoverReport> TaskCollection<T>.withVariant(variantName: String): TaskCollection<T>
fun <T : KoverReport> TaskCollection<T>.withMainVariant(): TaskCollection<T>
```
?

shanshin: It's a good idea, but there are few small drawbacks:
- this code will not work in old Groovy scripts (and many projects are still on Groovy) and in all the examples you will need to write that this code is only for kts
- these functions are still not very discoverable, if users are used to writing `matching`, then they do not know in advance about the existence of special `withVariant` functions (even if we call it `withKoverVariant`).

fzhinkin (approved):

PR #604, opened 2024-05-03
Added project filter for reports

### LINKED ISSUE #584: Ability to create custom report variant for specific dependency project

I would like to speed up my build by aggregating some tests into a single module. At the same time, I would like to retain the ability to check coverage of each module independently.

Imagine a project with 3 subprojects: `:app`, `:core`, and `:lib`. Let's say that the way things are now is that each subproject adds the kover plugin, and has its own tests. However, the overhead of running starting 3 separate test processes is too much. So, I want to only have one test test in `:app`, which will also test `:core` and `:lib` (3 projects is not significant overhead, but in real life there could be hundreds of subprojects).

I am looking at the newest API for creating custom variants in 8.0.0-Beta. I see in `KoverVariantCreateConfig` that I can create a new variant, however it doesn't seem that a variant can define its own dependencies. So here is the issue. Let's say `:app` adds `kover` depedencies on `:core:` and `:lib`. My understanding is that by default, all 3 modules code will be combined for the coverage verification rules. This means that you might be able to pass a rule even though you didn't cover a single line in `:lib`, because you covered enough lines in `:core`. 

I was hoping custom variants solve this problem, but I cannot figure out how to do what I am imagining. I would like a custom variant in `:app` in which verification rules are only looking at a single kover dependency project, like `:core`. 

Possibly the closest thing I am finding looking at the API is `KoverVariantSources.excludedSourceSets`. If only there was something like `KoverVariantSources.includedKoverDependencies: Set<String>` in which each string was the path of a module. If empty, it could default to including all kover dependencies as well as the current project. but if we were to write `includedKoverDependencies.append(':core')` then that variant would analyze its rules only using the code fom `:core`, not even from the current project.

shanshin:
Resolves #584


sandwwraith (approved):

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverReportsConfig.kt
```diff
@@ -672,6 +676,19 @@ public interface KoverReportFilter {
...
 
+    /**
+     * Add all classes in specified project. Only the project path is used (starts with a colon).
+     *
+     * It is acceptable to use `*` and `?` wildcards,
+     * `*` means any number of arbitrary characters (including no chars), `?` means one arbitrary character.
+     *
+     * Example:
+     * ```
+     *  projects.add(":my:project-path")
```
sandwwraith, line 687: Perhaps you can also add example that shows that wildcards are also working instead of `:`, e.g. `projects.add("*libs")`

PR #625, opened 2024-06-07
Partially apply recommendations for improving DSL

### LINKED ISSUE #600: Improve DSL by leveraging Gradle managed objects
# What is your use-case and why do you need this feature?
I recently put focus into this project while I was researching it for a potential replacement on JaCoCo.

While exploring `0.8.0-Beta2` usage and internals (source code), I've identified some pain points in the current experimental DSL, that IMHO should be addressed by leveraging Gradle-managed objects, like `NamedDomainObjectContainer` and its variants. 

I didn't scan fully what the project has to offer, like `filtering`, etc. So you may consider this as partial feedback.

# Describe the solution you'd like

## Child closures in DSL
The main `kover: KoverProjectExtension` extension has several child closures, like `reports: KoverReportsConfig`.

These closures can only be accessed by calling a method (i.e. `kover.reports { }`).

As my current understanding of Gradle's patterns, child closures should be accessible by property or method closure, so the following two syntaxes should be valid and equivalent:
```kotlin
kover.reports.doSomething()
// and
kover.reports {
   doSomething()
}
```

This can easily be achieved by changing a bit the DSL object definition, i.e.:
```kotlin
interface KoverProjectExtension {

    val reports: KoverReportsConfig
    
    fun reports(action: Action<KoverReportsConfig>) = action.configure(reports)

}
```

## `kover.variants` DSL
Kover currently supports (as far I noted) three types of variants: `provided`, `customer`, and `total`

It would be nice if you iterate the DSL to be able to access all declared variants with a `NamedDomainObjectSet` (at least), or even better, use a `PolymorphicDomainObjectContainer` to also use Gradle's standard API for creating them.

For instance, having:
```kotlin
interface KoverProjectExtension {
    val variants: PolymorphicDomainObjectContainer<KoverVariant>
}

interface KoverVariantsContainer : PolymorphicDomainObjectContainer<KoverVariant> {
    val provided: NamedDomainObjectCollection<KoverProvidedVariant>
    val custom: NamedDomainObjectCollection<KoverCustomVariant>
    val total: KoverTotalVariant
}

// minimal DSL interfaces
interface KoverVariant: Named
interface KoverProvidedVariant : KoverVariant
interface KoverCustomVariant : KoverVariant
interface KoverTotalVariant : KoverVariant
```

This could be a basic implementation for `KoverVariantsContainer`:
```kotlin
internal abstract class KoverVariantsContainerImpl @Inject constructor(
    objects: ObjectFactory,
) : KoverVariantsContainer,
    PolymorphicDomainObjectContainer<KoverVariant> by (objects.polymorphicDomainObjectContainer(KoverVariant::class).apply {
        // only `KoverCustomVariant`s can be registered through DSL
        registerBinding(KoverCustomVariant::class.java, KoverCustomVariant::class.java)
    }) {

    override val provided = withType<KoverProvidedVariant>()

    override val custom = withType<KoverCustomVariant>()

    override val total = objects
        .newInstance<KoverTotalVariant>("total")
        .also(::add)

    internal fun addProvidedVariant(name: String, action: Action<KoverProvidedVariant>) = objects
        .newInstance<KoverProvidedVariant>(name)
        .also(action::execute)
        .also(::add)
}
```

You could have a richer DSL, supporting many use cases out of the box:

#### Iterating all variants in a hot collection
```kotlin
kover.variants.configureEach {
    sources.excludedSourceSets += "main"
}
```

#### Configuring variant in a generic way
Imaging #599 is implemented in a way that `KoverVariant` now has an `aggregate` property.

We can target all `debug` variants at once with:
```kotlin
kover.variants.provided.configureEach {
    aggregate = name.lowercase().endsWith("debug")
}
```

#### Simplified plugin internal wiring
Currently, your plugin has some manual checks to prevent variants of different types from collishing with others with the same name.

If you centralize that into a root `NamedDomainObjectContainer`, this problem will be gone (less code).

Later the plugin can just react on others, like Android one, and automatically register `provided` ones:
```kotlin
val extension = project.extensions.create<KoverProjectExtension>("kover")

androidComponents.onVariants variant@{
    val koverVariant = extension.variants.addProvidedVariant(this@variant.name) {
       // configure koverVariant
    }
}

```

shanshin:
The new DSL will allow to write chains like `kover.reports.total { ... }` instead of `kover { reports { total { ... } } }`.

However, the DomainObjectContainer for variants has not been implemented, because it will be necessary to rework the workflow with different types of variants, which is part of a larger reworking of the DSL.

Resolves #600

### DIFF
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverProjectExtension.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverProjectExtension.kt
@@ -81,6 +81,13 @@ public interface KoverProjectExtension {
     public fun currentProject(block: Action<KoverCurrentProjectVariantsConfig>)
 
     /**
+     * Instance to configuring of report variants shared by the current project.
+     *
+     * See details in [currentProject].
+     */
+    public val currentProject: KoverCurrentProjectVariantsConfig
+
+    /**
      * Configuration of Kover reports.
      *
      * An individual set of reports is created for each Kover report variant.
@@ -119,6 +126,13 @@ public interface KoverProjectExtension {
      * ```
      */
     public fun reports(block: Action<KoverReportsConfig>)
+
+    /**
+     * Instance to configuring of Kover reports.
+     *
+     * See details in [reports].
+     */
+    public val reports: KoverReportsConfig
 
     /**
      * Configuring a merged report.
@@ -200,6 +214,13 @@ public interface KoverProjectExtension {
      * ```
      */
     public fun merge(block: Action<KoverMergingConfig>)
+
+    /**
+     * Instance to configuring a merged report.
+     *
+     * See details in [merge].
+     */
+    public val merge: KoverMergingConfig
 
     @Deprecated(
         message = "Function excludeJavaCode was removed, to exclude all Java sources write here `currentProject { sources { excludeJava = true } }` or `currentProject { sources { excludeJava.set(true) } } instead. Please refer to migration guide in order to migrate: ${KoverMigrations.MIGRATION_0_7_TO_0_8}",
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/internal/MergeImpl.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/internal/MergeImpl.kt
@@ -17,25 +17,31 @@ internal abstract class KoverMergingConfigImpl: KoverMergingConfig {
     internal var instrumentationAction: Action<KoverMergingInstrumentation>? = null
     internal val variantsAction: MutableMap<String, Action<KoverMergingVariantCreate>> = mutableMapOf()
 
+    internal var configured: Boolean = false
 
     override fun subprojects() {
         subprojectsFilters += Spec<Project> { true }
+        configured = true
     }
 
     override fun subprojects(filter: Spec<Project>) {
         subprojectsFilters += filter
+        configured = true
     }
 
     override fun allProjects() {
         allProjectsFilters += Spec<Project> { true }
+        configured = true
     }
 
     override fun allProjects(filter: Spec<Project>) {
         allProjectsFilters += filter
+        configured = true
     }
 
     override fun projects(vararg projectNameOrPath: String) {
         allProjectsFilters += Spec<Project> { project -> project.name in projectNameOrPath || project.path in projectNameOrPath }
+        configured = true
     }
 
     override fun sources(config: Action<KoverMergingVariantSources>) {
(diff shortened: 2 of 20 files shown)


kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverProjectExtension.kt
```diff
@@ -77,6 +77,13 @@ public interface KoverProjectExtension {
      */
     public fun currentProject(block: Action<KoverCurrentProjectVariantsConfig>)
 
+    /**
+     * Instance to configuring of report variants shared by the current project.
+     *
+     * See details in [currentProject].
```
sandwwraith, line 83: Since KDoc doesn't have any syntax for overloads, I'm afraid that `[currentProject]` will link recursively to this property again (it doesn't do this in IDEA currently, but who knows). Maybe just say `See details in [currentProject] function` or `See details in [KoverCurrentProjectVariantsConfig]`?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/internal/MergeImpl.kt
```diff
@@ -17,46 +17,55 @@ internal abstract class KoverMergingConfigImpl: KoverMergingConfig {
     internal var instrumentationAction: Action<KoverMergingInstrumentation>? = null
     internal val variantsAction: MutableMap<String, Action<KoverMergingVariantCreate>> = mutableMapOf()
 
+    internal var configured: Boolean = false
```
sandwwraith, line 20: I think some test is needed for this new boolean property, with e.g., `kover.merge.subprojects()` in configuration

shanshin: It's already tested, see tests in `MergingTests`, e.g. `testRootNoProjects`

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverProjectExtension.kt
shanshin: In dokka when you click on the `currentProject` property, the function opens %)

So, it seems to me that this postscript is more for the IDE.

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/internal/MergeImpl.kt
sandwwraith: I mean you need to test that syntax `kover.merge.subprojects` works the same as `kover { merge { subprojects() } }`

sandwwraith (approved):

PR #632, opened 2024-06-12
Added test on copyVariant function

shanshin:


kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/ExamplesBuildTests.kt
```diff
@@ -24,6 +24,11 @@ internal class ExamplesBuildTests {
         // build only
     }
 
+    @ExamplesTest("jvm/copy-variant", ["koverHtmlReportFirst", "koverHtmlReportSecond"])
+    fun CheckerContext.jvmCopyVariant() {
+        // build only
```
sandwwraith, line 29: Why build only? Shouldn't you check that `first` variant covers only `SubprojectUtils` and `ClassFromSecondProject` is absent?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/FinalizeKover.kt
```diff
@@ -58,13 +58,6 @@ internal fun KoverContext.finalizing(origins: AllVariantOrigins) {
     jvmVariant?.let { variantArtifacts[JVM_VARIANT_NAME] = it }
     androidVariants.forEach { variantArtifacts[it.variantName] = it }
 
-    val availableVariants = variantArtifacts.keys + projectExtension.currentProject.customVariants.keys
```
sandwwraith, line 61: Is this related to the test? I don't know why this code got deleted

shanshin: This code was accidentally added during the merging, but the error was not detected because there was no test

sandwwraith (approved):

Issue #633, opened 2024-06-14
koverCachedVerify failure due to incorrect paths in the .artifact file


remcomokveld:
**Describe the bug**

Sporadically, in a medium-to-large project which has a couple of included builds, the `koverCachedVerify` of some android modules will fail due to having inputs from tasks that it didn't declare.

The error below is the result, of the bug, but I think the bug is further upstream in the Gradle plugin.

When I look at the files in `included-build/project-a/build/kover/.artifact`, `included-build/project-a/build/kover/debug.artifact` and `included-build/project-a/build/kover/release.artifact` they all contain the following

```
:project-b
project-b/src/debug/kotlin
project-b/src/debug/java
project-b/src/main/kotlin
project-b/src/main/java

project-b/build/tmp/kotlin-classes/debug
project-b/build/intermediates/javac/debug/compileDebugJavaWithJavac/classes

project-b/build/kover/bin-reports/testDebugUnitTest.ic
```

**Errors**

I guess as a result of the `.artifact` files having the wrong paths in them, the following gradle error appears.

```shell
* What went wrong:
A problem was found with the configuration of task ':included-build:project-a:koverCachedVerify' (type 'KoverDoVerifyTask').
  - Gradle detected a problem with the following location: '/Users/me/Projects/included-build/project-b/build/kover/bin-reports/testDebugUnitTest.ic'.

    Reason: Task ':included-build:project-a:koverCachedVerify' uses this output of task ':included-build:project-b:testDebugUnitTest' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.

    Possible solutions:
      1. Declare task ':included-build:project-b:testDebugUnitTest' as an input of ':included-build:project-a:koverCachedVerify'.
      2. Declare an explicit dependency on ':included-build:project-b:testDebugUnitTest' from ':included-build:project-a:koverCachedVerify' using Task#dependsOn.
      3. Declare an explicit dependency on ':included-build:project-b:testDebugUnitTest' from ':included-build:project-a:koverCachedVerify' using Task#mustRunAfter.

    For more information, please refer to https://docs.gradle.org/8.7/userguide/validation_problems.html#implicit_dependency in the Gradle documentation.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.
```

**Expected behavior**
[...]

PR #636, opened 2024-06-14
Fix issue with identical cache keys between projects

remcomokveld:
Fixes https://github.com/Kotlin/kotlinx-kover/issues/633

### DIFF
--- a/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/ArtifactGenerationTests.kt
+++ b/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/ArtifactGenerationTests.kt
@@ -26,6 +26,20 @@ internal class ArtifactGenerationTests {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


remcomokveld:
Actually, a good point was raised [here](https://gradle-community.slack.com/archives/CA83B1VLL/p1718420209744209?thread_ts=1718398398.983589&cid=CA83B1VLL) that this task should probably not be cacheable,  and I'm actually doubtful whether this task should exist at all. I don't think it would ever be quicker to write and then read this data from disk compared to just recomputing it every time in the places where this is used.

Especially on projects with a remote build cache where, in the absence of a valid output a network request would be done to check for remote entries.

See docs here https://docs.gradle.org/current/userguide/build_cache.html#sec:task_output_caching_disabled_by_default

Thoughts on either disabling caching or completely removing the task?

shanshin:
Hi, thanks for the effort!

We have thought about various options and at the moment we decided to try to disable the build cache for this task.

However, the behavior should be fixed, i.e. it should work with the build cache enabled. Therefore, we should  merge this PR and later #637.

shanshin (approved):

PR #637, opened 2024-06-17
Disable caching on kover artifact tasks

remcomokveld:
Fixes https://github.com/Kotlin/kotlinx-kover/issues/633

Alternative solution to https://github.com/Kotlin/kotlinx-kover/pull/636

Caching should only be enabled for tasks that do heavy computations and this task does not do so.

Potentially it would be good to merge both PRs, even if caching is disabled, we should make sure that data that is used to produce an output is declared as an import.

### DIFF
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tasks/services/KoverArtifactGenerationTask.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tasks/services/KoverArtifactGenerationTask.kt
@@ -11,6 +11,7 @@ import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.provider.*
 import org.gradle.api.tasks.*
 import org.gradle.kotlin.dsl.*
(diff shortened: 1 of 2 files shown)


kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tasks/services/KoverArtifactGenerationTask.kt
```diff
@@ -19,7 +19,6 @@ import javax.inject.*
  *
  * This artifact that will be shared between projects through dependencies for creating merged reports.
  */
-@CacheableTask
```
shanshin, line 22: Let's specify `@DisableCachingByDefault` instead of `@CacheableTask`

shanshin, line 22: Also, please, rerun all tests after this change, `BuildCacheRelocationTests` must fall.

shanshin (approved):
Thanks!

Issue #608, opened 2024-05-15
Design proposal: move the current functionality of Kover Gradle Plugin to Kotlin Gradle Plugin


shanshin:
# Abstract
Calculating code coverage for applications written in Kotlin is a basic functionality closely related to the language itself.

Now this functionality is supplied separately in the form of a Gradle plugin, which is in no way related to the Kotlin compiler used.
Enabling coverage measurement requires a number of additional actions, using a plugin, and configuring scripts through extensions unrelated to Kotlin.

The Kotlin coverage settings have nothing to do with the Kotlin settings, which makes these settings less obvious.

For a smoother use of Kotlin, we decided to move the DSL coverage collection settings to Kotlin Gradle Plugin.

# Motivation
Disadvantages of the current approach:
- the complexity of the concepts used in Kover Gradle Plugin: it introduces its own concepts of artifacts, coverage measurement options, report variants, its own rules for combining variants, its own rules for merging reports. All these concepts do not fit well with Kotlin, and its look like an artificial pile-up
- after creating a minimal Kotlin application, it is not obvious that you need to use a third-party utility to measure coverage
- some Kover plugin settings duplicate existing settings, for example, the need to add an `kover` dependency when a `implementation` dependency has already been declared - this complicates the configuration for builds with a large number of subprojects
- unobvious to configure Kover as a separate Gradle plugin that is not related to Kotlin in any way
- difficulty of supporting two Gradle plugins in the IDE - it is easier to support one Kotlin Gradle Plugin than two plugins with different versions
- the current plugin DSL is specific to Java or JaCoCo in many respects
- complicated integration with the Kotlin compiler
- different lifecycle of Kover Gradle Plugin and Kotlin Gradle Plugin potentially complicates delivering support for new compiler features

Among the additional advantages, integration with Kotlin Gradle Plugin will allow to add the functionality of native targets instrumentation more smoothly

# Continued development of Kover Gradle Plugin
We will continue to maintain the current plugin and also add some small but convenient functionality. 

Planned changes:
- fix new critical bugs in Kover Gradle Plugin
- fix new bugs in the JVM agent
- [improve DSL by leveraging Gradle managed objects](https://github.com/Kotlin/kotlinx-kover/issues/600)
- fix problem with [JaCoCo logging](https://github.com/Kotlin/kotlinx-kover/issues/544) or if it is not possible to specify this behavior in the documentation
[...]

PR #644, opened 2024-06-24
Implemented protype of Kover Aggregated Plugin

shanshin:
Relates #608


kover-gradle-plugin/build.gradle.kts
```diff
@@ -87,7 +108,7 @@ val functionalTest by tasks.registering(Test::class) {
         systemProperties["junit.jupiter.execution.parallel.config.fixed.parallelism"] =
             junitParallelism?.toIntOrNull()?.toString() ?: "2"
         // this is necessary if tests are run for debugging, in this case it is more difficult to stop at the test you need when they are executed in parallel and you are not sure on which test the execution will pause
-        systemProperties["junit.jupiter.execution.parallel.enabled"] = if (junitParallelism == "no") "false" else "true"
+        systemProperties["junit.jupiter.execution.parallel.enabled"] = if (junitParallelism == "no") "false" else "false"
```
fzhinkin, line 111: false-false

shanshin: `#define true false`

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/KoverSettingsGradlePlugin.kt
```diff
@@ -0,0 +1,117 @@
...
+            group = "verification"
+            includedProjects.convention(settingsExtension.reports.includedProjects)
+            excludedProjects.convention(settingsExtension.reports.excludedProjects)
+            excludedClasses.convention(settingsExtension.reports.excludedClasses)
+            includedClasses.convention(settingsExtension.reports.includedClasses)
+            title.convention(projectPath)
+
+            htmlDir.convention(layout.buildDirectory.dir("reports/kover/html"))
+
+            this.onlyIf {
```
fzhinkin, line 87: Is there a reason to use `onlyIf` for logging? Won't `doFirst/doLast` suit this purpose better?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/KoverSettingsGradlePlugin.kt
```diff
@@ -0,0 +1,117 @@
...
+            dependsOn(artifacts)
+            this.artifacts.from(artifacts)
+            group = "verification"
+            includedProjects.convention(settingsExtension.reports.includedProjects)
+            excludedProjects.convention(settingsExtension.reports.excludedProjects)
+            excludedClasses.convention(settingsExtension.reports.excludedClasses)
+            includedClasses.convention(settingsExtension.reports.includedClasses)
+            title.convention(projectPath)
+
+            htmlDir.convention(layout.buildDirectory.dir("reports/kover/html"))
```
fzhinkin, line 85: Should the path constant be placed in `KoverPaths`?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/KoverParametersProcessor.kt
```diff
@@ -0,0 +1,44 @@
...
+package kotlinx.kover.gradle.aggregation.settings
+
+import kotlinx.kover.gradle.aggregation.settings.dsl.intern.KoverSettingsExtensionImpl
+import org.gradle.api.provider.HasMultipleValues
+import org.gradle.api.provider.ProviderFactory
+
+
+internal object KoverParametersProcessor {
+    fun process(settingsExtension: KoverSettingsExtensionImpl, providers: ProviderFactory) {
+        if (providers.gradleProperty("kover").isPresent) {
```
fzhinkin, line 14: Does it make sense to check support values for this property? For instance, suppose someone has a project where coverage was enabled using `enableCoverage` DSL's call, but a user now want's to explicitly disable it via CLI by passing `-Pkover=false`.

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/KoverParametersProcessor.kt
```diff
@@ -0,0 +1,44 @@
...
+            settingsExtension.coverageIsEnabled.set(true)
+        }
+
+        settingsExtension.reports.includedProjects.readAppendableArgument(providers, "kover.classes.from")
```
fzhinkin, line 18: I guess `kover.classes.from` should reads as "collect coverage for classes from following projects", but for me options like `kover.classes.from.excludes` and `kover.classes.excludes` are not easily distinguishable.
Did you consider naming package-related options as `kover.packages.includes` and `kover.packaged.excludes`?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/KoverParametersProcessor.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    private fun HasMultipleValues<String>.readAppendableArgument(providers: ProviderFactory, propertyName: String) {
+        val propertyProvider = providers.gradleProperty(propertyName)
+        if (propertyProvider.isPresent) {
+            val arg = propertyProvider.get().parseCollection()
+            if (!arg.append) {
+                empty()
+            }
+            println("Property '$propertyName' = ${arg.values}")
```
fzhinkin, line 31: Should it be printed through info/debug logging instead?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/KoverParametersProcessor.kt
```diff
@@ -0,0 +1,44 @@
...
+            addAll(arg.values)
+        }
+    }
+
+    private fun String.parseCollection(): ArgCollection {
+        val append = startsWith('+')
```
fzhinkin, line 37: Just curious, is this notation widespread among Gradle plugins?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/tasks/KoverXmlReportTask.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.TaskAction
+
+@CacheableTask
+internal abstract class KoverXmlReportTask : AbstractKoverTask() {
+    @get:OutputFile
+    internal abstract val reportFile: RegularFileProperty
+
+    @get:Input
+    abstract val title: Property<String>
```
fzhinkin, line 22: The title seems to be shared by both tasks, maybe move it to the abstract one?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/dsl/intern/KoverAggregatedExtensionImpl.kt
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright 2017-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.kover.gradle.aggregation.settings.dsl.intern
+
+import org.gradle.api.provider.SetProperty
+
+internal abstract class KoverAggregatedExtensionImpl {
```
fzhinkin, line 9: What's the purpose of this class? It seems to be unused

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/project/tasks/KoverAgentSearchTask.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.gradle.api.GradleException
+import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.tasks.*
+import org.gradle.work.DisableCachingByDefault
+
+/**
+ * Task to get online instrumentation agent jar file by specified coverage tool.
+ *
+ * The task is cached, so in general there should not be a performance issue on large projects.
```
fzhinkin, line 17: > The task is cached

> @DisableCachingByDefault

🤔

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/project/KoverProjectGradlePlugin.kt
```diff
@@ -0,0 +1,181 @@
...
+        val pluginManager = pluginManager
+        val projectPath = path
+
+        val compilations = project.layout.buildDirectory.map {
+            val compilations = when {
+                pluginManager.hasPlugin(KOTLIN_JVM_PLUGIN_ID) -> {
+                    val kotlin = exts.findByName("kotlin")?.bean()
+                        ?: throw KoverCriticalException("Kotlin JVM extension not found")
+                    kotlin["target"]["compilations"].sequence().filter { compilation ->
+                        compilation["name"].value<String>() != "test"
```
fzhinkin, line 85: It'll be filtered the second time below, so maybe we can skip filtration here.

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/commons/utils/Dynamic.kt
```diff
@@ -0,0 +1,54 @@
...
+    operator fun get(name: String): DynamicBean = getNotNull("get property '$name'").getProperty(name).bean()
+
+    operator fun contains(name: String): Boolean = getNotNull("check for a property '$name'").hasProperty(name)
+
+    inline fun <reified T> value(): T {
+        val notNull = origin ?: throw IllegalStateException("Value is null, failed to get value")
+        return notNull as? T ?: throw IllegalStateException("Invalid property value type, expected ${T::class.qualifiedName}, found ${notNull::class.qualifiedName}")
+    }
+
+    inline fun <reified T> orNull(): T? {
```
fzhinkin, line 23: unused

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/project/KoverProjectGradlePlugin.kt
```diff
@@ -0,0 +1,181 @@
...
+            .map { dir -> testTasks.map { dir.file(binReportName(it.name)) } }
+
+        val exts = extensions
+        val pluginManager = pluginManager
+        val projectPath = path
+
+        val compilations = project.layout.buildDirectory.map {
+            val compilations = when {
+                pluginManager.hasPlugin(KOTLIN_JVM_PLUGIN_ID) -> {
+                    val kotlin = exts.findByName("kotlin")?.bean()
```
fzhinkin, line 82: Nice trick with the DynamicBean!

kover-gradle-plugin/build.gradle.kts
fzhinkin: Jokes aside, why parallel execution is now always disabled?

kover-gradle-plugin/docs/aggregated.md
```diff
@@ -0,0 +1,67 @@
...
+If no tests were called in the assembly, then the coverage for all classes will be 0.
+
+## Configuring
+At the moment, Kover Settings Plugin allows to configure reports minimally.
+
+There are two ways to configure, using a build script in `settings.gradle.kts` and CLI
+
+Acceptable settings in `settings.gradle.kts` and their equivalent in CLI
+```kotlin
+kover {
```
fzhinkin, line 35: If the end goal is moving this plugin to KGP, perhaps we should abstain from using `kover` as an extension name and use something less interesting, but at the same time also less obscure. For instance, `coverage` or `codeCoverage`. WDYT?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/KoverSettingsGradlePlugin.kt
shanshin: Because `onlyIf` always executed, even when the task does not start and the result is taken from the cache (`FROM-CACHE` outcome)

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/KoverParametersProcessor.kt
shanshin: The names are not final at all and this is the first thing that came to mind. 
In any case, you can come up with a lot of spelling options and they are all likely to be similar to each other in one way or another (

There will most likely be no packages - this filter is duplicated classes, and instead of `kover.classes.from`, we can try `kover.projects` and `kover.projects.excludes`

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/KoverParametersProcessor.kt
shanshin: No, but I do not know if there is a generally accepted syntax for such functionality.

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/tasks/KoverXmlReportTask.kt
shanshin: In addition to HTML and XML reports, there are others that lack a title, so I don't see the point in creating a class because of one property

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/dsl/intern/KoverAggregatedExtensionImpl.kt
shanshin: Yes, it's useless now.

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/project/tasks/KoverAgentSearchTask.kt
shanshin: Oops )

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/project/KoverProjectGradlePlugin.kt
shanshin: Thanks

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/commons/utils/Dynamic.kt
shanshin: This is a reserve for the future)

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/project/KoverProjectGradlePlugin.kt
shanshin: I don't like dynamic objects (
But this is the easiest way to get around the problems with different class loaders.

kover-gradle-plugin/build.gradle.kts
shanshin: I forgot to change it back, it makes it easier to connect the debugger.

kover-gradle-plugin/docs/aggregated.md
shanshin: So far, the purpose of this plugin is to collect feedback and test ideas for compatibility with Gradle.

Now we can't even guess what it will look like in KGP, what DSL it will be and what names will be used, because for this we need to connect additional teams.

It was enough for me so far that `kover {` and `-Pkover` write shorter than `coverage {` )

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/KoverParametersProcessor.kt
fzhinkin: > the users can write their own condition

That requires committing some code to a repo, so if someone want to temporary disable code coverage collection in CI, it could be problematic.

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/KoverSettingsGradlePlugin.kt
fzhinkin: It's worth plunking a comment describing it then

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/KoverParametersProcessor.kt
shanshin: in this case, we can add only `-Pkover=false`, so that there is `-Pkover` to turn on and `-Pkover=false` to turn off

kover-gradle-plugin/docs/aggregated.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Quickstart
+This plugin is a prototype for testing ideas related to an alternative configuration method.
+
+The main difference from the existing Kover Gradle Plugin is the reactive content of the report: only those classes and tests that were compiled and executed within the same build get into the report.
+
+To use the plugin, just add into a `settings.gradle.kts` file 
+```kotlin
+plugins {
+    id("org.jetbrains.kotlinx.kover.settings") version "0.8.1"
```
sandwwraith, line 11: I'm not sure that `kover.settings` conveys the meaning of the plugin. Yes, it is applied in settings, but its goals are 1) reduce configuration 2) instrument only test tasks that are currently run. We may want to eventually rename it, but I don't have any good name right now except `kover.onthefly`

kover-gradle-plugin/docs/aggregated.md
```diff
@@ -0,0 +1,67 @@
...
+    
+    reports {
+        // -Pkover.projects.includes=:a
+        includedProjects.add(":a")
+        
+        // -Pkover.projects.excludes=:b
+        excludedProjects.add(":b")
+
+        // -Pkover.classes.includes=classes.to.exclude.*
+        includedClasses.add("classes.to.include.*")
```
sandwwraith, line 47: I believe there should also be inclusions/exclusions by annotation

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/dsl/intern/KoverSettingsExtensionImpl.kt
```diff
@@ -0,0 +1,36 @@
...
+
+@Suppress("LeakingThis")
+internal abstract class KoverSettingsExtensionImpl @Inject constructor(
+    objects: ObjectFactory
+) : KoverSettingsExtension {
+    abstract val coverageIsEnabled: Property<Boolean>
+
+    override val reports: ReportsSettings = objects.newInstance<ReportsSettings>()
+
+    //    abstract override val variants: NamedDomainObjectContainer<VariantI>
```
sandwwraith, line 23: Not needed anymore?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/settings/dsl/intern/KoverProjectExtensionImpl.kt
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright 2017-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.kover.gradle.aggregation.settings.dsl.intern
+
+internal abstract class KoverProjectExtensionImpl {
```
sandwwraith, line 7: This class seems unnecessary

kover-gradle-plugin/build.gradle.kts
```diff
@@ -199,3 +215,18 @@ gradlePlugin {
...
+
+gradlePlugin {
+    website.set("https://github.com/Kotlin/kotlinx-kover")
+    vcsUrl.set("https://github.com/Kotlin/kotlinx-kover.git")
+
+    plugins {
+        create("KoverSettingsPlugin") {
+            id = "org.jetbrains.kotlinx.kover.settings"
+            implementationClass = "kotlinx.kover.gradle.aggregation.settings.KoverSettingsGradlePlugin"
+            displayName = "Gradle Plugin for Kotlin Code Coverage Tools"
```
sandwwraith, line 227: It has the same `displayName` and `description` as a regular plugin. Won't it be confusing?

kover-gradle-plugin/docs/aggregated.md
sandwwraith: Given that we have `merge` in regular setup which also does some kind of aggregation and various `AggregationType`s, I'm afraid that this word is already overused :(

shanshin: In fact, this plugin has nothing to do with old Kover plugin, but it does exactly what `merge` says - it creates a aggrgated reports

sandwwraith: Well, if I run `./gradlew :projectA:jvmTest -Pkover koverHtmlReport`, it will show me only projectA's coverage, in that case there's no aggregation

kover-gradle-plugin/build.gradle.kts
shanshin: I changed it a little

kover-gradle-plugin/docs/aggregated.md
shanshin: Not quite. `:projectA` may depends on `:projectB`, thus the sources from these two projects will be aggregated in the report

sandwwraith: I can live with `aggregation` for now, maybe we'll find better name later

PR #655, opened 2024-07-14
Include Android compilations in aggregation

ephemient:
The types are different but they look the same through `DynamicBean`.


shanshin (approved):
Many thanks for effort!

Indeed, this plugin has started to be developed for the JVM, although it will be most convenient for Android projects.

shanshin:
@ephemient, released in `0.8.3`

ephemient:
@shanshin Thank you for merging (and fixing) and releasing! The aggregation prototype has basic usability for us already. I'd like to see how verification will work though :)

PR #668, opened 2024-08-07
Replaced JaCoCo ant-calls with programmatic calls of JaCoCo's classes

### LINKED ISSUE #630: Replace JaCoCo ant-calls with programmatic calls of JaCoCo's classes
At the moment, ant-calls from the `org.jacoco.ant` dependency are used to work with JaCoCo.

This is a less reliable way to use dependencies, because we can only check the correctness of the code in runtime, we have to use the AntBuilder mechanism [that we cannot control](https://github.com/Kotlin/kotlinx-kover/issues/606), [unwanted code is executed](https://github.com/Kotlin/kotlinx-kover/issues/544), and it is also difficult to add new functionality to an existing.

Changes:
- calling the JaCoCo code inside Gradle workers with 
- using classes from `org.jacoco.report` to work with JaCoCo, e.g. `RulesChecker` for verification
- analyze if its API changes for different versions of JaCoCo, then we will need to create your own module for each subsequent version with an individual dependency on the corresponding version of JaCoCo 
- suggestion: allow users to specify their own `CoverageTool` heirs so that their coverage tools can be implemented

Related issues, should be fixed by this changes: https://github.com/Kotlin/kotlinx-kover/issues/606, https://github.com/Kotlin/kotlinx-kover/issues/544

### LINKED ISSUE #666: Different behaviour between Kover & JaCoCo coverage tools
When I use Kover without Jacoco I have information about line coverage:
`Task :app:koverPrintCoverageDevRelease
:app line coverage: 18.0846%`
But when I add: 
` kover {
        useJacoco()
    }`
I always get:
`Task :app:koverPrintCoverageDevRelease
:app line coverage: 0%
[ant:jacocoReport] Rule violated for bundle :app: lines covered ratio is 0.0000, but expected minimum is 1.0000`

I try to use filters but it wasn't help. It isn't work with filters and without one.

Environment:
Kover Gradle Plugin version: 0.7.6 and 0.8.3
Gradle version: 8.1
Kotlin project type: Kotlin/Android
Coverage Toolset: Kover, JaCoCo

shanshin:
Resolves #630
Fixes #666


kover-features-jvm/src/main/java/kotlinx/kover/features/jvm/KoverLegacyFeatures.kt
```diff
@@ -197,11 +197,12 @@ public object KoverLegacyFeatures {
...
-
-                rule.violations.forEach { bound ->
-                    messageBuilder.append("  ")
-                    messageBuilder.appendLine(bound.format(rule))
-                }
+                rule.violations.map { bound -> bound.format(rule) }
+                    .toSortedSet()
+                    .forEach { bound ->
+                        messageBuilder.append("  ")
+                        messageBuilder.appendLine(bound.format(rule))
```
sandwwraith, line 204: Didn't you already sort it in `.map { bound -> bound.format(rule) }`?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tools/jacoco/PrintCoverage.kt
```diff
@@ -0,0 +1,70 @@
...
+import kotlinx.kover.gradle.plugin.util.ONE_HUNDRED
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.provider.Property
+import org.gradle.workers.WorkQueue
+import java.io.File
+import java.math.BigDecimal
+
+internal fun ReportContext.printJacocoCoverage(request: CoverageRequest, outputFile: File) {
+    val bound = VerificationBound(ONE_HUNDRED, BigDecimal.ZERO, request.metric, request.aggregation)
+    val failRule = VerificationRule(true, "", request.entity, listOf(bound))
```
sandwwraith, line 25: Hm, is this the rule that we apply by default if none are specified?

kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/VerificationTests.kt
```diff
@@ -97,14 +97,14 @@ internal class VerificationTests {
 
         run("koverHtmlReport", "koverVerify", errorExpected = true) {
             verification {
-                assertKoverResult("""Rule 'counts rule' violated:
-  lines covered percentage is 46.590900, but expected minimum is 58
-  lines covered count is 41, but expected maximum is 3
+                assertResult("""Rule 'counts rule' violated:
+  lines covered count is *, but expected maximum is 3
```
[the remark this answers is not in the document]
shanshin: `*` means "any number". 
Because the percentage of instruction coverage may vary slightly between compiler versions. This test does not check specific values, so it is allowed to skip the actual coverage value

kover-features-jvm/src/main/java/kotlinx/kover/features/jvm/KoverLegacyFeatures.kt
```diff
@@ -197,11 +197,12 @@ public object KoverLegacyFeatures {
                 messageBuilder.appendLine("$namedRule violated: ${rule.violations[0].format(rule)}")
             } else {
                 messageBuilder.appendLine("$namedRule violated:")
-
-                rule.violations.forEach { bound ->
-                    messageBuilder.append("  ")
-                    messageBuilder.appendLine(bound.format(rule))
-                }
+                rule.violations.map { bound -> bound.format(rule) }
+                    .toSortedSet()
```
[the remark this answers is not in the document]
shanshin: This ensures that the same error message is guaranteed to always be the same for the same verification errors, regardless of the order in which these errors are found (the iteration order may differ in Kover and JaCoCo)

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tools/jacoco/PrintCoverage.kt
shanshin: No, this is a rule that will never passed (min=100, max=0).
Thus, JaCoCo will inform us of the current coverage value in error message.

Since JaCoCo doesn't just have an API for explicitly obtaining coverage values, we get them indirectly through verification.

sandwwraith: > Since JaCoCo doesn't just have an API for explicitly obtaining coverage values, we get them indirectly through verification.

I think this info is worth adding as a comment

sandwwraith (approved):

PR #674, opened 2024-08-19
Applied Simple Kotlin DSL Property Assignment

niyajali:
- https://blog.gradle.org/simpler-kotlin-dsl-property-assignment
- Enable Typesafe Project Accessors

### DIFF
--- a/kover-jvm-agent/build.gradle.kts
+++ b/kover-jvm-agent/build.gradle.kts
@@ -32,8 +32,8 @@ koverDocs {
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
+    sourceCompatibility = JavaVersion.VERSION_1_6
+    targetCompatibility = JavaVersion.VERSION_1_6
 }
 
 repositories {
--- a/kover-jvm-agent/build.gradle.kts
+++ b/kover-jvm-agent/build.gradle.kts
@@ -23,17 +23,17 @@ plugins {
 }
 
 koverPublication {
-    description.set("Kover JVM instrumentation agent")
+    description = "Kover JVM instrumentation agent"
 }
 
 koverDocs {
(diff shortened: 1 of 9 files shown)


kover-jvm-agent/build.gradle.kts
```diff
@@ -23,17 +23,17 @@ plugins {
...
-    docsDirectory.set("jvm-agent")
-    description.set("Kover JVM instrumentation agent")
+    docsDirectory = "jvm-agent"
+    description = "Kover JVM instrumentation agent"
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_6
-    targetCompatibility = JavaVersion.VERSION_1_6
+    sourceCompatibility = JavaVersion.VERSION_1_8
```
shanshin, line 35: We are specifically compiling for the `1.6` version for wider support of different JVMs.
Therefore, it is better to leave `1.6`

kover-offline-runtime/build.gradle.kts
```diff
@@ -23,12 +23,12 @@ plugins {
...
 
 koverPublication {
-    description.set("Compiled dependency to ensure the operation of the code that has been instrumented offline")
+    description = "Compiled dependency to ensure the operation of the code that has been instrumented offline"
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_7
-    targetCompatibility = JavaVersion.VERSION_1_7
+    sourceCompatibility = JavaVersion.VERSION_1_8
```
shanshin, line 30: It is also better to leave `1.7` here

kover-jvm-agent/build.gradle.kts
niyajali: But unable to run `clean build` locally, it says `error: Source option 6, 7 is no longer supported. Use 8 or later. `
I reverted back to as it is before, let see CI Build results.

shanshin (approved):
Thanks for the effort!

PR #680, opened 2024-08-26
Added USAGE attribute to all Kover configurations

### LINKED ISSUE #678: Gradle Plugin: Configurations should add Gradle.USAGE attribute to eg kover
**Describe the bug**
All outgoing configurations of Kover don't set the Gradle.USAGE attribute. Without this attribute, Gradles attribute matching algorithm accepts kovers artifacts even if you (as Gradle plugin author) want to resolve artifacts with other attributes.

Use-case: I applied the kover plugin using the settings plugin and have a custom Gradle plugin that wants to resolve a configuration with USAGE but there is no configuration with matching USAGE so Gradle should fail. But Gradles attribute matching algorithm (wrongly) choose a consumable configuration as long as there are no conflicting attributes, and ignores requires/provided attributes. So without USAGE attribute, Gradle simply choose kover artifacts. The Gradle USAGE attribute is the most relevant attribute (`Indicates main purpose of variant`)  and Gradle authors should set it.

**Errors**
n/a

**Expected behavior**
A Gradle variant error because I fetch Gradle.USAGE="foo" and kover provides Gradle.USAGE="kover"

**Reproducer**
n/a

**Reports**
n/a

**Environment**
 - Kover Gradle Plugin version: 0.8.3
 - Gradle version: 8.10
 - Kotlin project type: Kotlin/JVM 2.0.20
 - Coverage Toolset: Kover
 - Other context important for this bug: n/a

shanshin:
Fixes #678

### DIFF
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/commons/artifacts/KoverUsageAttr.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/commons/artifacts/KoverUsageAttr.kt
@@ -1,0 +1,13 @@
+/*
+ * Copyright 2017-2024 JetBrains s.r.o. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.kover.gradle.aggregation.commons.artifacts
+
+import org.gradle.api.attributes.*
+
+internal interface KoverUsageAttr: Usage {
+    companion object {
+        const val VALUE = "kover"
+    }
+}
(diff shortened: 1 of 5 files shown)


kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/LoggingTaskTests.kt
```diff
@@ -45,7 +45,7 @@ internal class LoggingTaskTests {
         run(":koverLog") {
             checkOutcome("koverPrintCoverage", "SUCCESS")
             taskOutput("koverPrintCoverage") {
-                assertEquals("application line coverage: 57.1429%\n\n", this)
+                assertContains(this, "application line coverage: 57.1429%\n\n")
```
sandwwraith, line 48: Is this change related? Why changing artifact attribute adds something to `koverPrintCoverage` output?

sandwwraith (approved):

kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/LoggingTaskTests.kt
shanshin: On some Gradle versions function `taskOutput` works incorrectly (has extra output). 
Will change it separately.

sandwwraith (approved):

PR #682, opened 2024-08-28
Made class KoverSettingsGradlePlugin for aggregated plugin public

shanshin:


sandwwraith:
Can you provide an example why is it needed?

shanshin:
> Can you provide an example why is it needed?

To apply plugin by syntax `project.pluginManager.apply(KoverSettingsGradlePlugin::class.java)`

sandwwraith (approved):

PR #688, opened 2024-09-18
Maven plugin - correction to documentation of filters

ollieabbey:
Spotted when setting up the plugin from the [docs](https://kotlin.github.io/kotlinx-kover/maven-plugin/).

`filters` shouldn't have another child block called `filters`.

Examples are already correct [here](https://github.com/Kotlin/kotlinx-kover/blob/main/kover-maven-plugin/examples/filters-common/pom.xml#L101) and [here](https://github.com/Kotlin/kotlinx-kover/blob/main/kover-maven-plugin/examples/filters-rules/pom.xml#L120).


shanshin (approved):
Thanks!

PR #726, opened 2025-01-07
Added ability to specify inclusion filter for source sets

### LINKED ISSUE #714: Explicitly included sourceSets (whitelist)
In my project, there's multiple test and development source sets, like `testIntegration` and `dev` (for dev environment setup code). Manually excluding all of them without overlooking ones is a pain. This is further amplified by the fact that the `testIntegration` source set is added using a `JvmTestSuite` convention that is only applied *after* Kover, therefore requiring the use of `afterEvaluate` in order to actually catch all non-main source sets.

Instead of just getting `excludedSourceSets` (aka "blacklist"), I'd like to have a "whitelist" approach as well. This way, I can only include the `main` source set and don't have to worry about the initialization order, dev source sets, ...

shanshin:
Resolves #714

### DIFF
--- a/kover-gradle-plugin/docs/index.md
+++ b/kover-gradle-plugin/docs/index.md
@@ -1120,6 +1120,8 @@ By specifying an empty filter `filters { }`, you can completely disable report f
 
 ### Exclusion of JVM source sets
 
+Code declarations from `test` source set are excluded by default.
+
 It is possible to exclude from all reports the code declared in certain source sets.
 
 ```kotlin
--- a/kover-gradle-plugin/docs/index.md
+++ b/kover-gradle-plugin/docs/index.md
@@ -1138,7 +1138,7 @@ In addition, it is possible to exclude classes from all source sets that are not
 kover {
     currentProject {
         sources {
-            excludedOtherSourceSets.addAll("main")
+            includedSourceSets.addAll("main")
         }
     }
 }
@@ -1151,7 +1151,7 @@ These filters can be combined, the `excludedSourceSets` filter has priority.
 kover {
     currentProject {
         sources {
-            excludedOtherSourceSets.addAll("main", "extra")
+            includedSourceSets.addAll("main", "extra")
             excludedSourceSets.addAll("main")
         }
     }
--- a/kover-gradle-plugin/docs/index.md
+++ b/kover-gradle-plugin/docs/index.md
@@ -1122,8 +1122,6 @@ By specifying an empty filter `filters { }`, you can completely disable report f
 
 It is possible to exclude from all reports the code declared in certain source sets.
 
-As a side effect, the generation of Kover reports ceases to depend on the compilation tasks of these source sets.
-
 ```kotlin
 kover {
     currentProject {
@@ -1133,6 +1131,35 @@ kover {
     }
 }
 ```
+
+In addition, it is possible to exclude classes from all source sets that are not specified.
+
+```kotlin
+kover {
+    currentProject {
+        sources {
+            excludedOtherSourceSets.addAll("main")
+        }
+    }
+}
+```
+
+This way, only classes from the `main` source set will be included in the report, but not from others.
+
+These filters can be combined, the `excludedSourceSets` filter has priority.
+```kotlin
+kover {
+    currentProject {
+        sources {
+            excludedOtherSourceSets.addAll("main", "extra")
+            excludedSourceSets.addAll("main")
+        }
+    }
+}
+```
+In this case, only classes from the `extra` source set will be included in the report.
+
+As a side effect of the source set exclusion, the generation of Kover reports ceases to depend on the compilation tasks of excluded source sets.
 
 ### Exclusion of test tasks
 If some task does not test the coverage of application classes, 
--- a/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/SourceSetsTests.kt
+++ b/kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/SourceSetsTests.kt
@@ -39,7 +39,7 @@ internal class SourceSetsTests {
 
         template.kover {
             currentProject {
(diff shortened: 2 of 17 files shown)


kover-gradle-plugin/docs/index.md
```diff
@@ -1134,6 +1132,35 @@ kover {
...
+kover {
+    currentProject {
+        sources {
+            excludedOtherSourceSets.addAll("main")
+        }
+    }
+}
+```
+
+This way, only classes from the `main` source set will be included in the report, but not from others.
```
sandwwraith, line 1147: Isn't this `includedSourceSets` then?

sandwwraith: Or at least `excludeEverySourceSetExcept`

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/artifacts/JvmVariantArtifacts.kt
```diff
@@ -48,9 +49,26 @@ internal class JvmVariantArtifacts(
...
     }
 
+    private fun compilationIsExcluded(compilationName: String, variant: KoverVariantSources): Boolean {
+        if (compilationName in variant.excludedSourceSets.get()) {
+            return false
+        }
+
+        val included = variant.excludedOtherSourceSets.get()
+        // if `test` is not included so we need to check it
+        if (compilationName == "test" && "test" !in included) {
```
sandwwraith, line 63: Isn't it the same condition as `included.isNotEmpty() && compilationName !in included`?

Also, you can add a test case for `"test"` source set name.

kover-gradle-plugin/docs/index.md
shanshin: I would like to get away from naming `included`. On the other hand, such things are now called `included` in the entire DSL.

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/artifacts/JvmVariantArtifacts.kt
shanshin: Indeed, I was a little hasty to add a separate condition

kover-gradle-plugin/docs/index.md
shanshin: Ok, let it be `included` for now

sandwwraith: I'm afraid we can move from this naming only for all functions at once, when we'll migrate whole Kover plugin

kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/cases/SourceSetsTests.kt
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        val gradleBuild = template.generate()
+        val buildResult = gradleBuild.runWithParams(":koverXmlReport")
+
+        gradleBuild.createCheckerContext(buildResult).xmlReport {
+            assertTrue(buildResult.isSuccessful, "Build should be successful")
+            classCounter("kotlinx.kover.examples.sourcesets.MainClass").assertAbsent()
+            classCounter("kotlinx.kover.examples.sourcesets.FooClass").assertAbsent()
+            classCounter("kotlinx.kover.examples.sourcesets.ExtraClass").assertPresent()
```
sandwwraith, line 98: Add `classCounter("kotlinx.kover.examples.sourcesets.TestClasses")` to this test and all above

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/artifacts/JvmVariantArtifacts.kt
```diff
@@ -48,9 +49,25 @@ internal class JvmVariantArtifacts(
...
     }
 
+    private fun compilationIsExcluded(compilationName: String, variant: KoverVariantSources): Boolean {
+        if (compilationName in variant.excludedSourceSets.get()) {
+            return false
+        }
+
+        val included = variant.includedSourceSets.get()
+
+        if (included.isEmpty() && compilationName == "test") {
```
sandwwraith, line 63: Are tests excluded by default? Is it mentioned somewhere in the documentation?

kover-gradle-plugin/docs/index.md
shanshin: Yes, I took this into account

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/artifacts/JvmVariantArtifacts.kt
shanshin: Yes, `test` source set excluded by default.

kover-gradle-plugin/src/functionalTest/kotlin/kotlinx/kover/gradle/plugin/test/functional/framework/starter/Commons.kt
```diff
@@ -183,3 +183,17 @@ internal fun File.patchKoverDependency(koverVersion: String) {
...
+
+internal fun File.addKoverBlocks(koverBlocks: MutableList<(ScriptLanguage, String) -> String>) {
+    if (koverBlocks.isEmpty()) return
+
+    val language = if (name.endsWith(".kts")) ScriptLanguage.KTS else ScriptLanguage.GROOVY
+
+    val builder = StringBuilder()
+    koverBlocks.forEach { block ->
+        builder.appendLine()
+        builder.append(block(language, "8.12"))
```
sandwwraith, line 195: Is 8.12 a Gradle version? Shouldn't it be some constant?

sandwwraith (approved):

PR #754, opened 2025-08-05
Disabled caching for KoverAgentJarTask

### LINKED ISSUE #748: KoverAgentJarTask cached execution is slower than not cached
I ran some gradle build scans to measure build performance and noticed something strange: it seems that `KoverAgentJarTask` runs slower when it is FROM-CACHE.

Could you help me understand why this is happening? Also, what is the best way to make this task non-cacheable?

Thank you!

Image

shanshin:
There are two cases where this task works:
- for the Kover tool it just copies the jar file from dependencies to the build directory
- for the JaCoCo tool it unzips the file from dependency and copies its content to the build directory

For both these cases using build cache is uneffective, since local copying and uzipping are fast itself. But using build cache inflates it, and it runs slower because the build cache performs some number of additional calculations and archiving.

Fixes #748

### DIFF
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tasks/services/KoverAgentJarTask.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tasks/services/KoverAgentJarTask.kt
@@ -4,22 +4,23 @@
 
 package kotlinx.kover.gradle.plugin.tasks.services
 
-import kotlinx.kover.gradle.plugin.dsl.*
-import kotlinx.kover.gradle.plugin.tools.*
+import kotlinx.kover.gradle.plugin.tools.CoverageTool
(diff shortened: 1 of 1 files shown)


kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tasks/services/KoverAgentJarTask.kt
```diff
@@ -4,22 +4,23 @@
...
+import org.gradle.work.DisableCachingByDefault
+import javax.inject.Inject
 
 /**
  * Task to get online instrumentation agent jar file by specified coverage tool.
  *
  * The task is cached, so in general there should not be a performance issue on large projects.
  */
-@CacheableTask
+@DisableCachingByDefault(because = "Local file operations are faster then using build cache")
```
sandwwraith, line 23: Won't it break cache for the tasks that depend on this one?

sandwwraith (approved):

PR #770, opened 2025-10-17
Aligned filtering for JaCoCo reports with Kover reporting

shanshin:
Added support of filtering by annotations and ancestors to the wrapper of the JaCoCo reporter.

Upgraded the minimal and default JaCoCo version to `0.8.13`

### DIFF
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tools/jacoco/JacocoFiltering.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tools/jacoco/JacocoFiltering.kt
@@ -20,13 +20,7 @@ import java.io.File
  */
 internal fun CoverageBuilder.filter(filters: ReportFilters, classfiles: Collection<File>): CoverageBuilder {
     // No filters - return original builder
-    if (filters.excludesClasses.isEmpty()
-        && filters.excludesAnnotations.isEmpty()
-        && filters.excludeInheritedFrom.isEmpty()
-        && filters.includesClasses.isEmpty()
-        && filters.includesAnnotations.isEmpty()
-        && filters.includeInheritedFrom.isEmpty()
-    ) {
+    if (filters.isEmpty) {
         return this
     }
 
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/commons/Types.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/commons/Types.kt
@@ -85,7 +85,16 @@ internal data class ReportFilters(
     val excludeInheritedFrom: Set<String> = emptySet(),
     @get:Input
     val excludeProjects: Set<String> = emptySet()
-): Serializable
+) : Serializable {
+    @get:Internal
+    val isEmpty: Boolean
+        get() = excludesClasses.isEmpty()
+                && excludesAnnotations.isEmpty()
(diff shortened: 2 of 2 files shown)


kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tools/jacoco/Commons.kt
```diff
@@ -27,75 +27,13 @@ internal interface CommonJacocoParameters: WorkParameters {
...
     tempDir.set(context.tempDir)
     projectPath.convention(context.projectPath)
 }
 
-/**
- * Replaces characters `|` or `\` to `.` and remove postfix `.class`.
- */
-internal fun String.filenameToClassname(): String {
-    return this.replace(File.separatorChar, '.').removeSuffix(CLASS_FILE_EXTENSION)
+internal fun String.toBinaryName(): String {
```
fzhinkin: Should it be private and declared in JacocoFiltering.kt?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverVersions.kt
```diff
@@ -17,7 +17,12 @@ public object KoverVersions {
     /**
      * JaCoCo coverage tool version used by default.
      */
-    public const val JACOCO_TOOL_DEFAULT_VERSION = "0.8.12"
+    public const val JACOCO_TOOL_DEFAULT_VERSION = "0.8.13"
+
+    /**
+     * JaCoCo coverage tool minimal supported version.
+     */
+    public const val JACOCO_TOOL_MINIMAL_VERSION = "0.8.13"
```
fzhinkin: What dictates this restriction? At first glance, it seems like it's only missing `fragments` property that stops us from using earlier version. And if I got it correctly, we may ignore it for now.

fzhinkin (changes requested):

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/dsl/KoverVersions.kt
shanshin: Yes, and this way we can guarantee support for inline functions. 
I do not know the reasons why older versions of JaCoCo might be needed.

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tools/jacoco/JacocoFiltering.kt
```diff
@@ -0,0 +1,300 @@
...
+            if (methodIsIncluded(regexFilter, classFiltered, methodAnnotations)) {
+                newClassCoverage.addMethod(methodCoverage)
+            }
+        }
+        newClassCoverage.signature = classCoverage.signature
+        newClassCoverage.setInterfaces(classCoverage.interfaceNames)
+        newClassCoverage.superName = classCoverage.superName
+        newClassCoverage.sourceFileName = classCoverage.sourceFileName
+        if (classCoverage is ClassCoverageImpl) {
+            newClassCoverage.fragments = classCoverage.fragments
```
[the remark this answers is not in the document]
shanshin: You are right, `fragments` important at the moment of the first `.visitCoverage(newClassCoverage)` - at this moment coverage of inline method is modified. So we can remove these lines.

PR #787, opened 2026-01-23
Added full support of Android Gradle plugin 9.0.0

### LINKED ISSUE #784: AGP 9.0.0 changes on Variant API
**Describe the bug**
I have androidApp project that contains activity, along with Instrumented UI tests and screenshot tests, I was able to collect screenshot tests coverage prior to AGP 9.0.0, because of Variant API change you could not query variants properly. Android Application plugin seems to be effected.

**Errors**
Does not find application variant I was using before `mockDebug`

**Expected behavior**
Should count screenshot test into coverage but it could not find variant I am using

**Reproducer**
[multiplatform](https://github.com/droideveloper/poc_multiplatform) :weather:androidApp
toggle `add('mockDebug', optional = true)` to `false`

**Reports**
my coverage reduced couple of percent

**Environment**
 - Kover Gradle Plugin version: [0.9.4]
 - Gradle version: [9.3.0]
 - Kotlin project type: [Kotlin/Multiplatform + Android + ios]
 - Coverage Toolset (if customized in build script): [Kover]

### LINKED ISSUE #785: AGP 9.0.0: Variant-specific report tasks not generated, koverXmlReport has no test task dependencies
## Describe the bug

With Android Gradle Plugin 9.0.0, Kover does not generate variant-specific report tasks (e.g., `koverXmlReportGoogleplayPremiumDebug`). Additionally, the total `koverXmlReport` task has no dependencies on test tasks, meaning running it independently produces an empty coverage report.

## Expected behavior

According to the [Kover documentation](https://kotlin.github.io/kotlinx-kover/gradle-plugin/#kotlin-single-module-android-project):

> For each Gradle Build Variant, Kover creates its own **report variant** that have the same names.

And:

> Running one of these tasks automatically triggers the launch of all test tasks for `debug` build variant

**Expected:**
- Variant-specific tasks like `koverXmlReportVariantDebug` should exist
- `koverXmlReportGoogleplayPremiumDebug` should depend on `testVariantDebugUnitTest`
- `koverXmlReport` (total) should depend on all test tasks

## Actual behavior

**With AGP 9.0.0:**
- Variant-specific report tasks do NOT exist
- `koverXmlReport` has NO test task dependencies
- Running `koverXmlReport` independently produces an empty report (0% coverage)

**Verification:**

```bash
# Check dependencies of koverXmlReport
./gradlew :Scribd:koverXmlReport --dry-run

# Output shows only:
:Scribd:koverGenerateArtifact SKIPPED
:Scribd:koverXmlReport SKIPPED

# No test tasks are included as dependencies
```

**Empty report generated:**
```xml
<?xml version="1.0" ?>
<report name="Kover Gradle Plugin XML report for :android">
<counter type="INSTRUCTION" missed="0" covered="0"/>
<counter type="BRANCH" missed="0" covered="0"/>
<counter type="LINE" missed="0" covered="0"/>
<counter type="METHOD" missed="0" covered="0"/>
<counter type="CLASS" missed="0" covered="0"/>
</report>
```

**Comparison with AGP 8.x (main branch):**

On AGP 8.x, the variant-specific task correctly depends on test tasks:
```bash
./gradlew :Scribd:koverXmlReportVariantDebug --dry-run
# Output includes:
:Scribd:testVariantDebugUnitTest SKIPPED
```

## Workaround

Explicitly run the test task before generating the report:
```bash
./gradlew :Scribd:testVariantDebugUnitTest :Scribd:koverXmlReport
```

This works because Kover's agent still instruments code and collects coverage during test execution; the report task then uses this collected data.

## Environment

- **Kover Gradle Plugin version:** 0.9.4
- **Gradle version:** 9.3.0
- **Android Gradle Plugin version:** 9.0.0
- **Kotlin version:** 2.3.0
- **Kotlin project type:** Android Application with multiple product flavors and build types
- **Coverage Toolset:** Kover (default)

## Additional context

This may be related to #784 (AGP 9.0.0 changes on Variant API), but that issue describes a different symptom (custom variant creation with `add()` not working). This issue is specifically about:
1. Variant-specific report tasks not being generated at all
2. The total report task missing test task dependencies

shanshin:
Implemented actions:
- added support of all android plugins to Kover Gradle Plugin
- added support of all android plugins to Kover Settings Plugin
- Added tests on AGP 9.0.0
- Tests for AGP 8.x.x have been created separately
- The examples have been translated to use AGP 9.0.0 (except example using missingDimensionStrategy in flavour)
- Added AGP version limit in README.md

Fixes #784
Fixes #785

### DIFF
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/project/KoverProjectGradlePlugin.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/project/KoverProjectGradlePlugin.kt
@@ -20,6 +20,8 @@ import kotlinx.kover.gradle.plugin.commons.ANDROID_BASE_PLUGIN_ID
 import kotlinx.kover.gradle.plugin.commons.KOTLIN_ANDROID_PLUGIN_ID
 import kotlinx.kover.gradle.plugin.commons.KoverCriticalException
 import kotlinx.kover.gradle.plugin.commons.hasAndroid9WithKotlin
+import kotlinx.kover.gradle.plugin.commons.isJavaCompilerOutputDirectory
+import kotlinx.kover.gradle.plugin.commons.isKotlinCompilerOutputDirectory
 import org.gradle.api.Action
 import org.gradle.api.Plugin
 import org.gradle.api.Project
@@ -151,12 +153,10 @@ internal class KoverProjectGradlePlugin : Plugin<Project> {
                     // since AGP 9.0.0 classesDirs are empty, so we should get the compilation tasks output
                     val kotlinCompileTask = compilation.value<TaskProvider<Task>>("compileTaskProvider")
                     val javaCompileTask = compilation.valueOrNull<TaskProvider<Task>>("compileJavaTaskProvider")
-                    // assumption: Kotlin class-files are not placed in directories named 'classpath-snapshot' and 'cacheable'
-                    val kotlinOutputs = kotlinCompileTask.map { it.outputs.files.filter { file -> file.name.let { name -> name != "classpath-snapshot" && name != "cacheable" } } }
+                    val kotlinOutputs = kotlinCompileTask.map { it.outputs.files.filter { file -> file.isKotlinCompilerOutputDirectory() } }
 
                     if (javaCompileTask != null) {
-                        // assumption: Java compiler places class-files in directories named 'classes'
-                        val javaOutputs = javaCompileTask.map { it.outputs.files.filter { file -> file.name.endsWith("classes") } }
+                        val javaOutputs = javaCompileTask.map { it.outputs.files.filter { file -> file.isJavaCompilerOutputDirectory() } }
                         providers.provider { objects.fileCollection().from(kotlinOutputs, javaOutputs) }
                     } else {
                         kotlinOutputs
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/commons/Paths.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/commons/Paths.kt
@@ -46,3 +46,14 @@ internal fun coverageLogPath(variant: String): String {
 internal fun artifactFilePath(variant: String): String = "kover${separator}$variant.artifact"
 
 private val separator = File.separatorChar
+
+// assumption: Kotlin class-files are not placed in directories named 'classpath-snapshot' and 'cacheable'
+internal fun File.isKotlinCompilerOutputDirectory(): Boolean {
+    val fileName = name
+    return fileName != "classpath-snapshot" && fileName != "cacheable"
+}
+
+// assumption: Java compiler places class-files in directories named 'classes'
+internal fun File.isJavaCompilerOutputDirectory(): Boolean {
+    return name == "classes"
+}
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/locators/Android.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/locators/Android.kt
@@ -11,6 +11,8 @@ import kotlinx.kover.gradle.plugin.commons.AndroidBuildVariant
 import kotlinx.kover.gradle.plugin.commons.AndroidFallbacks
 import kotlinx.kover.gradle.plugin.commons.AndroidFlavor
 import kotlinx.kover.gradle.plugin.commons.KoverIllegalConfigException
+import kotlinx.kover.gradle.plugin.commons.isJavaCompilerOutputDirectory
+import kotlinx.kover.gradle.plugin.commons.isKotlinCompilerOutputDirectory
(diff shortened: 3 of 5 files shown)


kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/commons/artifacts/ArtifactSerializer.kt
```diff
@@ -95,7 +97,29 @@ internal class CompilationInfo(
     @get:PathSensitive(PathSensitivity.RELATIVE)
     val sourceDirs: Collection<File>,
 
+    @get:InputFiles
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    val outputDirs: Provider<out FileCollection>
+)
+
+internal class ProjectArtifactInfoDeserialized(
```
fzhinkin: Why do we need `*Deserialized` classes identical to the non-deserialized classes?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/project/KoverProjectGradlePlugin.kt
```diff
@@ -136,26 +136,33 @@ internal class KoverProjectGradlePlugin : Plugin<Project> {
...
-                    val javaCompileTask = compilation.value<TaskProvider<Task>>("compileJavaTaskProvider").orNull
-                    val kotlinOutputs = kotlinCompileTask.outputs.files.filter { file -> file.name == "classes" }.files
-
-                    outputDirs = if (javaCompileTask != null) {
-                        val javaOutputs = javaCompileTask.outputs.files.filter { file -> file.name == "classes" }.files
-                        kotlinOutputs + javaOutputs
+                    val kotlinCompileTask = compilation.value<TaskProvider<Task>>("compileTaskProvider")
+                    val javaCompileTask = compilation.valueOrNull<TaskProvider<Task>>("compileJavaTaskProvider")
+                    // assumption: Kotlin class-files are not placed in directories named 'classpath-snapshot' and 'cacheable'
+                    val kotlinOutputs = kotlinCompileTask.map { it.outputs.files.filter { file -> file.name.let { name -> name != "classpath-snapshot" && name != "cacheable" } } }
```
fzhinkin: Why it is necessary to filter out these directories?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/project/KoverProjectGradlePlugin.kt
```diff
@@ -136,26 +136,33 @@ internal class KoverProjectGradlePlugin : Plugin<Project> {
...
+
+                    if (javaCompileTask != null) {
+                        // assumption: Java compiler places class-files in directories named 'classes'
+                        val javaOutputs = javaCompileTask.map { it.outputs.files.filter { file -> file.name.endsWith("classes") } }
```
fzhinkin: Are there some `android-classes`? Why it's not enough to pick just `classes`?

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/commons/artifacts/ArtifactSerializer.kt
shanshin: Because serialized class contains Provider of Files, while during deserialization classes already resolved, and passing ProviderFactory from a project actually requires a lot more changes than adding a new class.

It's trade-off (

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/project/KoverProjectGradlePlugin.kt
shanshin: > Why it is necessary to filter out these directories?

Because I can't make sure that in certain cases these directories won't contain class-files.
Duplication of class-files causes direct errors in the reporter, as well as "floating" coverage.

> Removing filtration does not cause functional tests failures, btw.

Yes, because in simple cases there are no class-files there.
Exactly, but I am not sure that when incremental compilation is enabled, other class-files will not appear.
At the same time, checking the directory names costs almost nothing.

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/project/KoverProjectGradlePlugin.kt
shanshin: I think I left it here by accident :)

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/aggregation/commons/artifacts/ArtifactSerializer.kt
fzhinkin: Ah, got it - I didn't spot the `Provider` when made the diff visually (I mean, using my eyes :) )

PR #811, opened 2026-05-05
Fix 0% coverage for classes in com.android.* packages (#810)

### LINKED ISSUE #810: com.android.* namespace incorrectly excluded from JVM instrumentation — 0% coverage for AOSP-style system apps (com.android.provision, com.android.systemui)
When applying the Kover plugin to a Kotlin/Android module whose package namespace follows 
AOSP conventions (`com.android.*`, e.g. `com.android.provision`, `com.android.systemui`), 
all project classes report **0% coverage** despite unit tests executing successfully.

The root cause is that Kover automatically writes `exclude=com.android.*` into the 
generated `kover-agent.args` file. This exclusion is presumably intended to prevent 
instrumenting Android SDK stub classes from `android.jar`, but it also excludes 
user application code that happens to share the same package prefix.

The issue **persists when using `kover { useJacoco() }`** — even in JaCoCo mode, Kover 
still generates the same `kover-agent.args` with `exclude=com.android.*`.

**Errors**
No errors in `build/kover/errors`. Tests pass, binary `.ic` report is written, 
but all classes in `com.android.*` show 0% in the HTML report.

Generated `build/tmp/testAndroid11DebugUnitTest/kover-agent.args`:
```
report.file=.../testAndroid11DebugUnitTest.ic
exclude=android.*
exclude=com.android.*
exclude=jdk.internal.*
```

**Expected behavior**
Classes declared in the project's own `com.android.provision.*` (or any `com.android.*`) 
package should be instrumented by the JVM agent and coverage should reflect actual 
test execution.

**Reproducer**
Any Kotlin/Android module with `namespace = "com.android.provision"` (or any 
`com.android.*` namespace) and unit tests for classes in that package.

Minimal `build.gradle.kts`:
```kotlin
plugins {
    id("com.android.application")
    id("org.jetbrains.kotlin.android")
}

android {
    namespace = "com.android.provision"
}
```

Unit test exists and passes. Running `./gradlew koverHtmlReportDebug` produces a 
report where all `com.android.provision.*` classes show 0% coverage.

Attempted workarounds that do **not** fix the instrumentation:

1. `kover { reports { filters { includes { classes("com.android.provision.*") } } } }`  
   → classes appear in the HTML report, but remain at 0% (agent never instrumented them)

2. `kover { useJacoco() }`  
   → `kover-agent.args` is still generated with `exclude=com.android.*`

3. `kover { currentProject { instrumentation { excludedClasses.add(...) } } }`  
   → only adds more exclusions, no API available to remove built-in defaults

**Reports**
HTML report shows all `com.android.provision.*` classes with 0% instruction, branch, 
line, and method coverage while tests for those classes are present and pass .

**Environment**
- Kover Gradle Plugin version: 0.9.8
- Gradle version: 8.11.1
- Kotlin project type: Kotlin/Android (multi-variant: android11Debug / android14Debug)
- Coverage Toolset: Kover (default), also reproduced with `useJacoco()`
- Other: AOSP-style system app - package namespace `com.android.*` is required 
  by the Android platform; renaming the package is not feasible

mkonkel:
Kover unconditionally added `exclude=com.android.*` to the JVM agent args to protect against instrumenting Android SDK stub classes. This also excluded user application code in AOSP-style packages (e.g. com.android.provision.*, com.android.systemui.*) with no way to opt out.

The fix changes the default-exclusion logic so that when the user explicitly includes a sub-pattern of a built-in Android exclusion via `includedClasses` (e.g. "com.android.provision.*"), the conflicting broader default exclusion ("com.android.*") is dropped from the agent args. The agent's include-whitelist then takes effect, instrumenting only the user's own package while leaving unmatched com.android.* classes uninstrumented.

Existing behavior is fully preserved when `includedClasses` is empty.

Fixes https://github.com/Kotlin/kotlinx-kover/issues/810

### DIFF
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/instrumentation/JvmTestTaskConfigurator.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/instrumentation/JvmTestTaskConfigurator.kt
@@ -39,20 +39,24 @@ internal fun TaskCollection<Test>.instrument(
 
         // Excludes Android SDK stub classes by default to prevent instrumentation errors,
         // see https://github.com/Kotlin/kotlinx-kover/issues/89.
-        // However, if the user explicitly includes a sub-pattern of a default Android exclusion
-        // (e.g. "com.android.provision.*" inside "com.android.*"), that default exclusion is
-        // dropped so the agent's include-whitelist can take effect for the user's own package.
-        val androidClasses = setOf(
-            "android.*", "com.android.*",
-            // Excluded to prevent errors when instrumenting JVM internals (e.g. JaCoCo + Robolectric).
-            "jdk.internal.*"
-        )
-        val excludedClassesWithAndroid = current.instrumentation.excludedClasses.zip(current.instrumentation.includedClasses) { excludedSet, includedSet ->
-            val effectiveAndroid = androidClasses.filter { androidPattern ->
-                val prefix = androidPattern.removeSuffix(".*") + "."
-                includedSet.none { includePattern -> includePattern.startsWith(prefix) }
-            }.toSet()
-            excludedSet + effectiveAndroid
+        // Set the Gradle property 'kover.android.excludes.disable' to opt out of the
+        // android.* / com.android.* exclusions (e.g. for AOSP system apps).
+        val androidExcludesDisabled = project.providers
+            .gradleProperty("kover.android.excludes.disable")
+            .map { true }
+            .orElse(false)
+
+        val excludedClassesWithAndroid = current.instrumentation.excludedClasses.zip(androidExcludesDisabled) { excludedSet, disableAndroid ->
+            val defaultExclusions = if (disableAndroid) {
+                setOf("jdk.internal.*")
+            } else {
+                setOf(
+                    "android.*", "com.android.*",
+                    // Excluded to prevent errors when instrumenting JVM internals (e.g. JaCoCo + Robolectric).
+                    "jdk.internal.*"
+                )
+            }
+            excludedSet + defaultExclusions
         }
 
         val taskInstrumentationDisabled = koverDisabled.map {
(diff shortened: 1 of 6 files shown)


shanshin:
Hi, I would suggest a slightly different, simpler solution.
Add the Gradle property, and when you specify it, filters "android.*" and "com.android.*" will not be added to the excluded classes

mkonkel:
Thanks for the suggestion! 
I've updated the PR to use a Gradle property `kover.android.excludes.disable` instead. When set, the `android.*` and `com.android.*` exclusions are not added to the agent args

kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/appliers/instrumentation/JvmTestTaskConfigurator.kt
```diff
@@ -37,14 +37,26 @@ internal fun TaskCollection<Test>.instrument(
...
 
-        // Always excludes android classes, see https://github.com/Kotlin/kotlinx-kover/issues/89
-        val excludedClassesWithAndroid = current.instrumentation.excludedClasses.map { it +
+        // Excludes Android SDK stub classes by default to prevent instrumentation errors,
+        // see https://github.com/Kotlin/kotlinx-kover/issues/89.
+        // Set the Gradle property 'kover.android.excludes.disable' to opt out of the
+        // android.* / com.android.* exclusions (e.g. for AOSP system apps).
+        val androidExcludesDisabled = project.providers
+            .gradleProperty("kover.android.excludes.disable")
+            .map { true }
```
shanshin, line 46: Always `true` if any value is specified?
Then it's better to use `.isPresent`

mkonkel: Thanks for the suggestion! I looked into it, but `isPresent` can't be used safely here due to Gradle's configuration cache.

`Provider<T>.isPresent` is a terminal, eager operation. It resolves the provider immediately and returns a plain Boolean. That Boolean then gets captured inside the lambda passed to `.zip() / .map {}`, and when Gradle serialises the task graph for the configuration cache, it bakes in the resolved value.

This means: if a developer runs `./gradlew test without -Pkover.android.excludes.disable`, the cache entry is stored with false. On the next run with `-Pkover.android.excludes.disable`, Gradle replays the cached configuration. The configureEach block never re-executes, `.isPresent` is never called again, and the stale false is used. The property change has no effect.

`.map { true }.orElse(false)` is actually semantically identical to `.isPresent (any value → true, absent → false)`, but it keeps a lazy `Provider<Boolean>` all the way down the chain, which is what Gradle needs to correctly invalidate the cache when the property changes between runs.

shanshin (approved):

PR #816, opened 2026-06-17
Fix AGP 9 built-in Kotlin projects routed into the pre-9 variant path

### LINKED ISSUE #815: Kover 0.9.8 hits the removed applicationVariants API on AGP 9
Applying Kover 0.9.8 to a `com.android.application` module on AGP 9.2.1 fails at configuration with:

```
Could not get unknown property 'applicationVariants' for object of type com.android.build.gradle.internal.dsl.ApplicationExtensionImpl$AgpDecorated
```

`applicationVariants` was removed in AGP 9. The only code that touches it is `androidVariantsBefore9()`, and `locateKotlinAndroidVariants()` only calls that when `androidMajorVersion() < 9`. On AGP 9.2.1 that should be 9, so it looks like it's hitting the `?: 0` fallback (the function can only return 9 or 0), and `0 < 9` then routes into the pre-9 path.

The module uses Kotlin provided by AGP, so there's no `kotlin-android` plugin applied and detection goes through `hasAndroid9WithKotlin()`.

I couldn't get a clean local repro. It only shows up on CI with the configuration cache on, which makes me think `androidComponents.pluginVersion` isn't readable at the point the locator runs during config cache store. Filing anyway since it's consistent there.

Environment
- Kover 0.9.8
- AGP 9.2.1
- Gradle 9.5.1
- Android application module, Kotlin provided by AGP (no kotlin-android plugin)

FrancoisBlavoet:
## Summary

Fixes #815.

`locateKotlinAndroidVariants()` picks the variant-extraction path based on `androidMajorVersion()`:

```kotlin
val majorVersion = project.extensions.androidMajorVersion()
val origins = if (majorVersion < 9) {
    project.androidVariantsBefore9(androidExtension, kotlinTarget)  // reads applicationVariants
} else {
    project.androidVariants(androidExtension, variants, kotlinTarget)
}
```

`androidMajorVersion()` reflectively reads `androidComponents.pluginVersion.major` and falls back to `0` when the read returns null, so it can only return the real major or `0` — never an intermediate value:

```kotlin
internal fun ExtensionContainer.androidMajorVersion(): Int {
    val androidComponents = findByName("androidComponents")?.bean() ?: return 0
    return androidComponents.beanOrNull("pluginVersion")?.valueOrNull<Int>("major") ?: 0
}
```

When that read falls back to `0` on an AGP 9 project, `0 < 9` routes into `androidVariantsBefore9()`, which reads the `applicationVariants` property removed in AGP 9 and fails configuration:

```
Could not get unknown property 'applicationVariants' for object of type
com.android.build.gradle.internal.dsl.ApplicationExtensionImpl$AgpDecorated
```

We hit this intermittently on CI on a `com.android.application` module using AGP's built-in Kotlin — only under the configuration cache and not on every run, which points at the reflective `pluginVersion` read returning null at the moment the locator runs.

**Why not a capability check on `applicationVariants`:** branching on whether the `android` extension still exposes `applicationVariants`/`libraryVariants` isn't reliable. The failing build proves `hasProperty("applicationVariants")` returns `true` on the AGP 9 `AgpDecorated` extension (the error is raised by the subsequent `getProperty`, only reached after `"applicationVariants" in androidExtension` passes), so `hasProperty` can't detect the removed API.

**Fix:** a project using AGP's built-in Kotlin — no `kotlin-android` plugin but a `kotlin` extension, i.e. `hasAndroid9WithKotlin()` — is necessarily AGP 9+, since built-in Kotlin was introduced in AGP 9. Treat that as AGP 9+ regardless of the version read:

```kotlin
val isAgp9OrNewer = majorVersion >= 9 || project.hasAndroid9WithKotlin()
val origins = if (!isAgp9OrNewer) { ... }
```

This drops the dependency on the fragile `pluginVersion` read for the built-in-Kotlin case (the one reaching this locator via the `hasAndroid9WithKotlin()` branch in `ProvidedVariantsLocator`), and is a no-op when the version read works or for AGP 7/8 projects using the `kotlin-android` plugin.

## Verification

I could not run the full Gradle build locally — the `build-logic` build needs a JDK 8 toolchain I don't have provisioned; `:kover-gradle-plugin:compileKotlin` otherwise picks up the change. Relying on CI for the full build. Happy to add a functional test if you can point me at the AGP 9 test fixture setup — the trigger is timing-dependent, so I'd appreciate guidance on simulating the `0` fallback deterministically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/locators/KotlinAndroidLocator.kt
+++ b/kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/locators/KotlinAndroidLocator.kt
@@ -7,6 +7,7 @@ package kotlinx.kover.gradle.plugin.locators
 import kotlinx.kover.gradle.plugin.appliers.origin.AllVariantOrigins
 import kotlinx.kover.gradle.plugin.commons.KoverCriticalException
 import kotlinx.kover.gradle.plugin.commons.androidMajorVersion
+import kotlinx.kover.gradle.plugin.commons.hasAndroid9WithKotlin
 import kotlinx.kover.gradle.plugin.util.bean
 import org.gradle.api.Project
 
@@ -25,7 +26,9 @@ internal fun Project.locateKotlinAndroidVariants(variants: List<AndroidVariantIn
     val kotlinTarget = kotlinExtension["target"]
 
     val majorVersion = project.extensions.androidMajorVersion()
-    val origins = if (majorVersion < 9) {
+    // Built-in Kotlin implies AGP 9+, even when androidMajorVersion() can't read the version (#815).
+    val isAgp9OrNewer = majorVersion >= 9 || project.hasAndroid9WithKotlin()
+    val origins = if (!isAgp9OrNewer) {
         project.androidVariantsBefore9(androidExtension, kotlinTarget)
     } else {
         project.androidVariants(androidExtension, variants, kotlinTarget)


kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/locators/KotlinAndroidLocator.kt
```diff
@@ -25,7 +26,9 @@ internal fun Project.locateKotlinAndroidVariants(variants: List<AndroidVariantIn
     val kotlinTarget = kotlinExtension["target"]
 
     val majorVersion = project.extensions.androidMajorVersion()
-    val origins = if (majorVersion < 9) {
+    // Built-in Kotlin implies AGP 9+, even when androidMajorVersion() can't read the version (#815).
+    val isAgp9OrNewer = majorVersion >= 9 || project.hasAndroid9WithKotlin()
```
FrancoisBlavoet, line 30: `androidMajorVersion()` reflectively reads `androidComponents.pluginVersion.major` and returns its `?: 0` fallback when the read yields null — so it's either the real major or `0`, never an intermediate. On AGP 9 that `0` makes `majorVersion < 9` true and routes into `androidVariantsBefore9()`, which reads the `applicationVariants` property removed in AGP 9 and crashes configuration (#815). It's intermittent for us, only under the configuration cache.

A naive capability check (does the `android` extension still expose `applicationVariants`?) doesn't work: the crash proves `hasProperty("applicationVariants")` returns `true` on the AGP 9 `AgpDecorated` extension — the failure comes from the later `getProperty`. So `hasProperty` can't detect the removed API.

`hasAndroid9WithKotlin()` (a `kotlin` extension with no `kotlin-android` plugin) only happens on AGP 9+, since built-in Kotlin was introduced in AGP 9 — a reliable AGP-9 signal that sidesteps the fragile version read. No-op when the read works or for AGP 7/8 `kotlin-android` projects.

PR #823, opened 2026-08-12
Serialize in-process coverage aggregation

### LINKED ISSUE #822: Concurrent report/verify tasks of different projects cross-contaminate through shared intellij-coverage statics (KoverTool aggregates in-process without isolation)
**Describe the bug**

In a multi-project build, when report/verify tasks of **different projects** execute concurrently in one Gradle daemon, their aggregations cross-contaminate. A project's verify counts classes that only a **sibling project's** exclude list names, or produces coverage numbers underivable from its own class set.

Because `koverCachedVerify*` is `@CacheableTask`, the wrong verdict is a *successful* task output containing a violation message, so it is published to the build cache under a key computed from that task's own healthy inputs and replayed to later builds that never raced.

A minimal reproducer is attached ([kover-concurrent-aggregation-repro.zip](https://github.com/user-attachments/files/30969227/kover-concurrent-aggregation-repro.zip)) — 8 plain-JVM modules, contaminated in 8 of 8 runs.

**Root cause**

`KoverTool` calls `KoverLegacyFeatures` directly on the task's thread — no `workerExecutor`, no isolation:

```kotlin
override fun verify(rules: List<VerificationRule>, output: File, context: ReportContext) {
    val violations = KoverLegacyFeatures.verify(...)
```

The same holds for `koverHtmlReport` / `koverXmlReport` / `koverBinaryReport` / `koverLog` (`KoverHtmlOrXmlReport.kt`, `IcReport.kt`, `Evaluation.kt`). By contrast, the JaCoCo tool path submits its work through `workerExecutor.classLoaderIsolation { ... }`.

`KoverLegacyFeatures` drives `com.intellij.rt.coverage`, whose classes hold shared mutable static state:

```
com.intellij.rt.coverage.data.ProjectData:  public static ProjectData ourProjectData;   // non-final
com.intellij.rt.coverage.util.OptionsUtil:  public static boolean FIELD_INSTRUMENTATION_ENABLED,
                                            CONDY_ENABLED, CALCULATE_HITS_COUNT, ...    // non-final
```

All projects' Kover tasks share one classloader in the daemon, so concurrent aggregations race over the same statics.

Two factors make this easy to hit:

1. **With the configuration cache enabled, Gradle executes tasks in parallel even when `org.gradle.parallel=false`.** Users who believe they have serialized execution have not. Confirmed by logging `System.nanoTime()` and the thread name in `doFirst`/`doLast` of every Kover task: five different projects' tasks entered concurrently on distinct execution workers with `org.gradle.parallel=false` set.
2. The `@CacheableTask` verdict propagation described above, which spreads a single contaminated run to unrelated builds.

**Expected behavior**

Report/verify tasks of different projects must be safe to execute concurrently. Options, in what we believe is decreasing confidence:

1. **Serialize the aggregations internally.** The calls all run on daemon threads, so a JVM monitor around the `KoverLegacyFeatures` call sites in the Kover tool path would suffice — no public API change, no worker plumbing, no effect on the JaCoCo path.
2. **Remove the shared static state** from the aggregation path in intellij-coverage. The real repair, but a larger change in a different repository.
3. **Isolate the aggregation** via the worker API. Flagging this as *unverified* rather than recommending it: Gradle caches classloaders by classpath for `classLoaderIsolation`, and every project here shares one reporter classpath, so concurrent work actions may share the classloader and therefore the statics. `processIsolation` would certainly work but costs a forked JVM per report task.

On the JaCoCo comparison: switching the reproducer to `useJacoco()` makes it clean (0 of 6 runs contaminated), but we do not think that shows isolation is what protects it — `org.jacoco.core`'s `Analyzer`, `ExecutionDataStore` and `CoverageBuilder` are instance-based with no mutable global statics, so there is nothing there to contaminate.

**Reproducer**

Attached: [kover-concurrent-aggregation-repro.zip](https://github.com/user-attachments/files/30969227/kover-concurrent-aggregation-repro.zip). Eight plain-JVM Kotlin modules, no Android, no KMP. Each module `mN` has:

- `Covered.kt` — 20 lines, fully exercised by that module's tests
- `Uncovered.kt` — 40 lines, never called
- an exclusion naming **only its own** uncovered class, `com.repro.mN.UncoveredKt`
- a verification rule requiring 100% line coverage

With its own filters applied each module is 20/20 = 100% and passes. If an aggregation applies a **sibling's** filters, that module's own 40-line uncovered class is counted: 20/60 = **33.3333%**, and verify fails.

```bash
./gradlew clean
./gradlew :m1:koverVerify :m2:koverVerify :m3:koverVerify :m4:koverVerify \
          :m5:koverVerify :m6:koverVerify :m7:koverVerify :m8:koverVerify \
          --no-build-cache
```

**`--no-build-cache` is required to observe the bug.** Without it, `koverCachedVerify*` is restored FROM-CACHE after the first run, never aggregates, and everything passes — the build cache masks the defect. `clean` does not clear it. This produced two full rounds of false negatives for us before we noticed.

`./loop.sh 8` repeats the cycle and counts contaminated rounds; `./loop.sh 8 -PkoverLock=true` applies the workaround below from the same checkout.

| condition | outcome |
|---|---|
| all 8 modules, one invocation, `--no-build-cache` | **8 of 8 runs contaminated**, always exactly `33.3333%` |
| `:m1:koverVerify` alone | passes — the exclusion configuration itself is correct |
| all 8 modules with `-PkoverLock=true` | **0 of 8 runs contaminated** |

The failing module subset varies per run (`[m1 m2 m4 m5 m6 m7 m8]`, then `[m5 m8]`, then `[m1]`) — a race rather than a configuration error.

**Reports**

From the ~90-module Android/KMP build where this first surfaced, same versions:

- A module gated at 100% with one excluded, never-loaded class failed at exactly its unfiltered value — `79.518100` = 66/83 lines, the excluded class's 17 lines counted at 0% — while a sibling module in the same invocation filtered correctly. The counted class belonged to a sibling's exclude list.
- A module failed verification at `83.333300` while **its own `koverHtmlReport` task in the same build** reported 50% over the same class set. Both wrong; the true value, confirmed by running the module alone, is 100%. Two aggregations of one project disagreeing with each other is not explicable by any per-task input.
- Execution-time instrumentation (a `doFirst` logging the task's `filters` property) showed the failing task holding its complete, correct filter set — the loss is inside aggregation, not in task configuration or property state.
- In that job the two projects' verify tasks started 0.55 ms apart, their html tasks 0.9 ms apart.

**Environment**

- Kover Gradle Plugin version: 0.9.9 (intellij-coverage 1.0.765)
- Gradle version: 9.6.1, with `org.gradle.configuration-cache=true`, `org.gradle.configuration-cache.parallel=true`, `org.gradle.parallel=false` (ignored under the configuration cache)
- Kotlin project type: Kotlin/JVM (reproducer); Kotlin/Multiplatform + Android (production build)
- Coverage Toolset: Kover (default). `useJacoco()` is unaffected.
- Other context: reproduced on multi-core macOS and on 4-core linux-aarch64 CI runners

**Workaround**

```kotlin
abstract class KoverAggregationLock : BuildService<BuildServiceParameters.None>

val lock = gradle.sharedServices.registerIfAbsent("koverAggregationLock", KoverAggregationLock::class.java) {
    maxParallelUsages.set(1)
}
tasks.matching { task ->
    listOf("koverCachedVerify", "koverHtmlReport", "koverXmlReport", "koverBinaryReport", "koverLog")
        .any { task.name.startsWith(it) }
}.configureEach { usesService(lock) }
```

`koverGenerateArtifact*` only writes path lists, and `koverPrintCoverage*` only prints a pre-written file (`KoverPrintLogTask` just reads it) — neither reaches the aggregator, so neither needs the lock. The aggregating task behind the log report is `koverLog*` (`KoverFormatCoverageTask`, which calls `collectCoverage`).

eugenepopovychfido:
Fixes #822.

## Problem

`KoverTool` calls `KoverLegacyFeatures` directly on the task's thread, with no worker isolation — unlike `JacocoTool`, which submits its work through `WorkerExecutor.classLoaderIsolation`. The intellij-coverage classes behind that call keep process-wide mutable state (`ProjectData.ourProjectData`, the mutable `OptionsUtil` flags), and all projects' report tasks share one classloader in the daemon.

Gradle runs tasks of different projects concurrently whenever the configuration cache is enabled, regardless of `org.gradle.parallel`, so two projects can aggregate at the same time over that shared state. The observable result is a report or verification verdict computed with another project's filters or coverage data. Since `koverCachedVerify*` is `@CacheableTask`, the wrong verdict is a *successful* task output containing a violation message, so it is stored under a key derived from the task's own healthy inputs and replayed to builds that never raced.

## Change

Serialize every entry into the aggregator on one monitor — the five call sites in `tools/kover` (`verify`, `aggregateIc`, `evalCoverage`, `generateHtmlReport`, `generateXmlReport`). `violationMessage` is left alone: it formats an already-computed result and touches no shared state.

The calls all run on daemon threads, so a JVM lock is sufficient. No public API change, no configuration, no worker plumbing, and the JaCoCo path is untouched.

This is containment rather than a root fix — the deeper repair is removing the shared statics from the aggregation path in intellij-coverage, which is a larger change in another repository. Happy to pursue that instead if you'd prefer.

## Validation

A minimal reproducer (8 plain-JVM modules, each excluding only its own uncovered class, each gated at 100%) is attached to #822. A module aggregating with a sibling's filters counts its own 40-line uncovered class and lands on 20/60 = `33.3333%`.

Against the **released 0.9.9**, using an external `BuildService` with `maxParallelUsages = 1` to serialize the same tasks:

| condition | contaminated |
|---|---|
| 8 modules, one invocation, `--no-build-cache` | 8 / 8 runs |
| single module alone | 0 — the exclusion configuration itself is correct |
| 8 modules + serialization | 0 / 8 runs |

Against a **locally-built plugin from this branch's base**, same reproducer and harness:

| build | contaminated |
|---|---|
| unpatched | 3 / 4 runs |
| this branch | 0 / 6 runs |

`--no-build-cache` is required to observe any of this: otherwise `koverCachedVerify*` is restored FROM-CACHE, never aggregates, and everything passes.

Two caveats on the local build, for transparency. Kover pins JDK 8 across `build-logic`, `kover-gradle-plugin`, `kover-features-jvm` and `kover-jvm-agent`, and no JDK 8 was available on the machine, so the validation build used a JDK 25 toolchain with a JVM 17 target. Those build-file changes are local only and are not part of this branch. The A/B holds the toolchain constant across both halves, so the comparison stands, but this is not a byte-for-byte stock build — your CI will exercise it properly.

## No test

CONTRIBUTING asks for a unit test, and I haven't added one. A meaningful test here asserts mutual exclusion around an in-process aggregator, which needs a seam the plugin doesn't currently expose; anything weaker either just asserts the wrapping exists or is inherently timing-dependent. I'd rather follow your preference than guess — happy to add whatever shape you consider useful, or to restructure the fix if you'd like the lock somewhere more testable.


kover-gradle-plugin/src/main/kotlin/kotlinx/kover/gradle/plugin/tools/kover/Evaluation.kt
```diff
@@ -13,15 +13,17 @@ import java.io.File
...
-    val coverage = KoverLegacyFeatures.evalCoverage(
-        request.entity.convert(),
-        request.metric.convert(),
-        request.aggregation.convert(),
-        tempDir,
-        filters.toKoverFeatures(),
-        files.reports.toList(),
-        files.outputs.toList()
-    )
+    val coverage = synchronized(KoverAggregationLock) {
```
shanshin, line 16: @dkhalanskyjb, could there be some hidden problems with this approach?

Personally, I don't see any problems, but suddenly I'm missing something.

dkhalanskyjb: The only actual problem this could cause is a deadlock if one task under the lock awaits the result of another task running in another thread with the same lock. From my understanding of these tasks, this should never happen, they look independent of one another.

So, from the point of view of correctness, I'd say this looks good to go. It's weird to me that not all usages of the IntelliJ API are serialized—for example, why isn't `KoverLegacyFeatures.mergeIc` also under the lock?—but that's your call, I'm not familiar with the codebase.

Stylistically, there's potential for improvement. **tl;dr:** it looks like the simplest solution is to stick a `@Synchronized` annotation to `verify`+`generateXmlReport`+`generateHtmlReport`(+`mergeIc`?), and be done.

In detail:
- Locks should typically surround the smallest amount of code. Otherwise, threads waiting to take the lock will be postponed by a longer time than is necessary. In particular, pure computations need to be extracted from under the lock. For `ReportContext.printCoverage` in particular, only the `LegacyVerification.verify` call needs to be protected—the rest of the code is just building lists and manipulating data.
- In the same vein, lock objects typically need to stay private. `KoverAggregationLock` can live in `KoverLegacyFeatures.kt` simply as `private val koverAggregationLock = Any()`.
- If it turns out all `synchronized` calls are in the same object after the change—in our case, in `KoverLegacyFeatures`—it could be easier to mark its methods with `@Synchronized`. E.g.,
```kotlin
    @Synchronized
    public fun verify(
        rules: List<Rule>, tempDir: File, filters: ClassFilters, binaryReports: List<File>, classfileDirs: List<File>
    ): List<RuleViolations>
```
It has the same effect as a `synchronized(this) { ... }` around their bodies (maybe that's not the case for Kotlin objects if you also use `@JvmStatic` on their methods, I'm not sure—but you don't, so that's fine).

shanshin: I agree about moving synchronization block inside `KoverLegacyFeatures` - this will help to atomise critical sections and not to expose lock outside.

shanshin: > why isn't KoverLegacyFeatures.mergeIc also under the lock?

Also agree. It's better to add critical section even for this method because it can potentially work with a static field.

shanshin:
Closed in prior of #829.

Thanks for the effort!

PR #829, opened 2026-09-15
Fixed concurrency problem with Kover reporter

### LINKED ISSUE #822: Concurrent report/verify tasks of different projects cross-contaminate through shared intellij-coverage statics (KoverTool aggregates in-process without isolation)

shanshin:
The IntelliJ coverage APIs maintain a mutable state shared within a classloader and are not safe to invoke concurrently. Concurrent calls can therefore interfere with one another.

To eliminate this, synchronization was added to the data loading methods.

Fixes #822


kover-features-jvm/src/main/java/kotlinx/kover/features/jvm/KoverLegacyFeatures.kt
dkhalanskyjb, line 50: Doesn't this need to be synchronized, too?

dkhalanskyjb (approved):

kover-features-jvm/src/main/java/kotlinx/kover/features/jvm/KoverLegacyFeatures.kt
shanshin: Instrumentation doesn't load coverage data and does not use a shared state

fzhinkin (approved):

